PR #4, opened 2016-03-15
Bracket

MelizzaP:


elements/bulbs-poll/bulbs-poll.scss
```diff
@@ -39,11 +40,47 @@ bulbs-poll {
...
+  .bulbs-poll-image-answers {
+    display: flex;
+  }
+
   .bulbs-poll-answers {
     margin: 0;
     padding: 0;
   }
 
+  li.bulbs-poll-image-answer {
```
collin: Let's not be specific as to the tag in these selectors. I've been following class or tag in this file, but not both.

elements/bulbs-poll/bulbs-poll.scss
```diff
@@ -73,7 +114,44 @@ bulbs-poll {
     }
   }
 
-  .bulbs-poll-answer {
+  li.bulbs-poll-image-result:before {
```
collin: Did this not work in the general * reset block above?

elements/bulbs-poll/components/answers.js
```diff
@@ -1,16 +1,24 @@
 import React, { PropTypes } from 'react';
 import Answer from './answer';
+import ImageAnswer from './image-answer';
 import classnames from 'classnames';
 
 export default function Answers (props) {
-  let classNames = classnames('bulbs-poll-answers', {
+  let AnswerType = Answer;
+  let ListType = 'bulbs-poll-answers';
```
collin: `AnswerType` needed to be capitalized because react requires `<Component>` to be capitalized as a constructor.

Something like `listClassName` would be more descriptive here.

elements/bulbs-poll/components/results-list.js
```diff
@@ -7,12 +8,17 @@ export default function ResultsList (props) {
     winningAnswers,
     vote,
   } = props;
-
+  let ResultType = Result;
+  let ListType = 'bulbs-poll-results-list';
+  if(props.poll.data.answer_type === 'imageText') {
+    ResultType = ImageResult;
+    ListType = 'bulbs-poll-image-results-list';
```
collin: Same comment as above re: casing here.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bulbs-elements",
-  "version": "0.1.1",
+  "version": "0.2.4",
```
collin: Until we're ready to pull the trigger on a production deploy, let's stick to "X.Y.Z-rc#". It should be safe to roll back from X.Y.4 to X.Y.3, but if we did a bunch of release candidates in between we won't know what version to roll back to.

collin:
:+1: Good PR :)

We're getting a little bit of duplication between the Image and non-image answers and results. But let's wait until we have to write up a third type of question/result to see if it's actually the right pattern, or just coincidence.

elements/bulbs-poll/bulbs-poll.scss
MelizzaP: no :-(

PR #8, opened 2016-04-05
Add test scripts for convenience and to maintain idiomatic npm project configuration

daytonn:
This is common for npm packages. Even though we have a convention of using scripts, we can still leverage npm to maintain the [principal of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).


kand:
:+1:

collin:
:+1:

PR #9, opened 2016-04-05
Simplified store

collin:
Modified the Store to take a plain object schema instead of wrapping objects and functions in classes that more or less did nothing.

@daytonn should sit down with you to go over how this would effect `<campaign-display>`


elements/bulbs-poll/fields/selected-answer.test.js
```diff
@@ -14,17 +14,17 @@ describe('<bulbs-poll> SelectedAnswerField', function () {
     let { selectAnswer } = actions;
 
     it('selects first answer', function () {
-      let nextState = selectAnswer.invoke(null, answer1);
+      let nextState = selectAnswer(null, answer1);
```
daytonn: I really like this. It seems much more intuitive to simply call a function than have to prefix it with `invoke`

elements/bulbs-poll/fields/src.js
```diff
@@ -1,10 +1,10 @@
-import { Field, Action } from 'bulbs-elements/store';
-
-const SrcField = new Field({
+const SrcField = {
   initialState: '',
-  setSrc: new Action(function (state, src) {
-    return src;
-  }),
-});
+  actions: {
```
daytonn: This is great, much more straight-forward.

PR #10, opened 2016-04-05
squashed simplified store

collin:
@kand @daytonn 

* removes the `Field` class
* removes the `Action` class
* replaces `Store` subclasses with POJO field schema


daytonn:
👍

.eslintrc
```diff
@@ -196,7 +198,6 @@
     "space-before-function-paren": 0,
     "space-in-parens": [1, "never"],
     "space-infix-ops": 1,
-    "space-return-throw-case": 1,
```
collin: Actually, upon looking up this rule I don't know why I was breaking it, going to look into it:
http://eslint.org/docs/rules/space-return-throw-case

collin: Ah, that's right:

`Replacement notice: This rule was removed in ESLint v2.0 and replaced by keyword-spacing rule.`

These rules were removed.

collin:
Merging this for @daytonn to pull into his campaign-display tests.

PR #11, opened 2016-04-05
tests and clarity around state immutability

collin:


lib/bulbs-elements/store/store.js
```diff
@@ -1,6 +1,29 @@
...
+  }
+
+  unsubscribeComponent (component) {
+    let index = this.components.indexOf(component);
+    if (index > -1) {
+      this.components.splice(index, 1);
+    }
+  }
+}
+
```
collin: This did not change. It just moved to the top of the file.

collin:
@daytonn @kand 

Added another few tests and clarified some naming around the core store.

lib/bulbs-elements/store/store.js
```diff
@@ -42,13 +65,15 @@ function dispatchFunction (store, action, payload) {
     console.time(`dispatch ${action.type}`);
   }
 
-  let nextState = cloneObject(store.state);
-  let actionState = nextState[action.fieldName];
-  let actionNextState = cloneObject(actionState);
-  actionNextState = action(actionNextState, payload, store);
-  nextState[action.fieldName] = actionNextState;
+  let priorState = store.state;
+  let nextState = Object.assign({}, priorState);
```
collin: We do a shallow copy of the store's full state with `Object.assign`.

This way, we can do a strict equals check on `store.state === store.state`. Good for react to know that a component needs to re-render.

lib/bulbs-elements/store/store.js
```diff
@@ -42,13 +65,15 @@ function dispatchFunction (store, action, payload) {
...
+  let priorFieldValue = priorState[action.fieldName];
+  let nextFieldValue = cloneObject(priorFieldValue);
```
collin: We do a deep clone on the actual field being changed so the action can make any modifications.

This is a poor-man's `immutable.js`.

lib/bulbs-elements/store/store.js
```diff
@@ -42,13 +65,15 @@ function dispatchFunction (store, action, payload) {
...
+
+  nextState[action.fieldName] = action(nextFieldValue, payload, store);
```
collin: Here is where the action happens, we set the field of the `nextState` to the result of the action.

lib/bulbs-elements/store/store.js
```diff
@@ -42,13 +65,15 @@ function dispatchFunction (store, action, payload) {
...
+
   store.state = nextState;
-  deliverSubscriptions(store, nextState);
+  deliverSubscriptions(store);
```
collin: And deliver state changes to all components subscribed to this store.

collin:
@spra85

lib/bulbs-elements/store/store.js
```diff
@@ -1,6 +1,29 @@
 import cloneObject from 'clone';
 import iterateObject from 'bulbs-elements/util/iterate-object';
 
+export default class Store {
+  constructor (options) {
+    this.actions = wrapActions(options.schema, this);
+    this.state = collectInitialState(options.schema);
+    this.components = [];
+    if (false) {
```
spra85: ?

spra85:
Makes sense.

lib/bulbs-elements/store/store.js
collin: The logging is a little noisy and we haven't made a choice on how to control it.

lib/bulbs-elements/store/store.js
daytonn: Speaking of the poor man's immutable.js. Would actually using Immutable.js here be desirable?

kand: If we're gonna go down that route, I feel like we should just use redux too.

collin: 56K for immutable.js :(

daytonn: I can buy that, is there a way we can prevent deep references or throw an error or something?

collin: Have an example of what you mean by that?

daytonn:
```js
var bar = {
  baz: "baz",
};

var obj = {
  foo: "foo",
  bar: bar,
};

var cloned = Object.assign({}, obj);

bar.baz = "mutated";
cloned.bar.baz === "mutated"; // true
```

daytonn: I guess what I meant was should we put more protection around this. I feel like this is an easy way for bugs to slip in since we're providing not-quite immutable structures.

collin: It is possible for a naughty action to reach into the store to mutate another field's state.
This in particular should not be done:
```js
// An action of the foo field
FooField = {
  initialState: {},
  actions: {
    naughtyAction(state, argument,  store) {
      store.state.bar.baz = 'mutated';
    },
  },
};
```

daytonn: Yeah, I think we can self police this unless it bites us too much. We just have to keep in mind that we have minimal protection against pass-by-reference.

collin: The state argument passed into the action is a deep clone, so we're protected at the action level.

daytonn: Oh, then I think we should be fine. It's those pesky object references that end up being the bane of every js developer's existence.

collin: Yeah, this object copying here is two steps:

1. shallow copy entire store
1. deep copy the value of the field undergoing transition

This is good for reasons:

1. we don't incur the cost of copying the entire state for every action
2. react can optimize it's re-render check by comparing object identity.

    `store.state === store.state` simple check re-render components that have the full state as a prop

    `store.state.foo === store.state.foo`  simple check to re-render components that have a field as a prop

This is not perfect for reasons:

1. react doesn't get such a simple identity check for more deeply nested fields

But, bulbs-elements elements are expected to be small components/widgets with small trees of state, not large applications with huge forests of state. This should be kept in mind as we consider additions to the set of elements. So far I don't see anything we've come up with as going out of scope.

daytonn: agreed

collin:
@daytonn @kand @spra85 any objections to merge?

daytonn:
Do it 👍

PR #12, opened 2016-04-06
Fix lint errors on constant condition

daytonn:
@kand @collin This error is preventing me from cutting a release candidate


collin:
:+1:

kand:
:+1:

For a real implementation, would this be pulled from the env or something?

daytonn:
@kand Yeah we haven't figured out how it will ultimately be implemented. At this point it's very noisy and I think it's more a convenience thing.

collin:
Yeah, originally it was enabled in all dev builds.

It's nice for debugging, but definitely would be nice to figure out how to
keep it out of karma shell output/flag at runtime.

PR #17, opened 2016-04-07
Protect against missing urls and network errors

daytonn:
@kand @collin As soon as I added tests for the empty component, I found out that I do not need an empty component. This commit is essentially throwing an error if no campaign url is provided and preventing all the campaign-display tests from making network calls.


kand:
:+1:

PR #18, opened 2016-04-08
Need to include credentials for the request

daytonn:
@kand @collin


collin:
https://github.com/theonion/bulbs-elements/blob/master/elements/bulbs-poll/fields/poll.js#L51-L56

This needs to be passed to `makeRequest` rather than to the action itself.

daytonn:
D'oh that's what you get for trying to go too fast.

collin:
# These Thumbs?

PR #19, opened 2016-04-08
Refactor request implementation

daytonn:
@kand @collin I need some eyes on this. The tests pass but I want to make sure I understand all the plumbing.

### DIFF
--- a/elements/campaign-display/campaign-display.js
+++ b/elements/campaign-display/campaign-display.js
@@ -23,7 +23,7 @@ class CampaignDisplay extends BulbsElement {
   render () {
     return (
       <CampaignDisplayRoot
-          data={this.state}
+          {...this.state}
           display={this.props.display}
       />
     );
--- a/elements/campaign-display/components/campaign-display-image.js
+++ b/elements/campaign-display/components/campaign-display-image.js
@@ -3,13 +3,14 @@ import React, { PropTypes } from 'react';
 export default function CampaignDisplayImage (props) {
   return (
     <div className='campaign-display-image'>
-      <a href={props.url}>
-        <img src={props.image} />
+      <a href={props.clickthrough_url}>
+        <img src={props.image_url} alt={props.name} />
       </a>
     </div>);
 }
 
 CampaignDisplayImage.propTypes = {
-  image: PropTypes.string.isRequired,
-  url: PropTypes.string.isRequired,
+  clickthrough_url: PropTypes.string.isRequired,
+  image_url: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
 };
--- a/elements/campaign-display/components/campaign-display-root.js
+++ b/elements/campaign-display/components/campaign-display-root.js
@@ -6,11 +6,11 @@ export default function CampaignDisplayRoot (props) {
   let component;
   switch (props.display) {
   case 'name':
-    component = <CampaignDisplayName name={props.data.name}/>;
+    component = <CampaignDisplayName name={props.campaign.name}/>;
     break;
 
   case 'image':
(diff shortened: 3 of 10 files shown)


elements/campaign-display/campaign-display.js
```diff
@@ -23,7 +23,7 @@ class CampaignDisplay extends BulbsElement {
   render () {
     return (
       <CampaignDisplayRoot
-          data={this.state}
+          {...this.state}
```
kand: This seems like unnecessary indirection, can we make this `campaign={this.state.campaign}` instead? Or is there a specific reason to keep it as a generic unpacking?

elements/campaign-display/components/campaign-display-image.js
```diff
@@ -3,13 +3,14 @@ import React, { PropTypes } from 'react';
 export default function CampaignDisplayImage (props) {
   return (
     <div className='campaign-display-image'>
-      <a href={props.url}>
-        <img src={props.image} />
+      <a href={props.clickthrough_url}>
```
kand: Something I noticed in avclub, we might need additional attributes on this anchor tag, like stuff for the analytics manager. Is there some way to we can pass arbitrary attributes to this anchor?

If not, or it's too sloppy, I think we can refactor out any issues the analytics manager has with resolving its attributes if those attributes don't live on the elements themselves.

collin: which attributes?

So we couldn't get away with putting the attributes directly on the `<campagin-display>` as it is?

I think that would be ideal.

elements/campaign-display/campaign-display.js
daytonn: I suppose, it's really the same thing

elements/campaign-display/components/campaign-display-image.js
daytonn: I'd like to wait until there are actual requirements to inform this decision. I have no knowledge of any other attribute needs.

kand: Actually, nevermind, if there's a problem with the analytics manager here, we should fix it in the analytics manager and keep this clean.

elements/campaign-display/components/campaign-display-root.js
```diff
@@ -6,11 +6,11 @@ export default function CampaignDisplayRoot (props) {
   let component;
   switch (props.display) {
   case 'name':
-    component = <CampaignDisplayName name={props.data.name}/>;
+    component = <CampaignDisplayName name={props.campaign.name}/>;
     break;
 
   case 'image':
-    component = <CampaignDisplayImage name={props.data.image} />;
+    component = <CampaignDisplayImage {...props.campaign} />;
```
kand: Same as above, can we explicitly name the attributes being used?

daytonn: I suppose, but then I don't get to use the fancy new spread operator.

elements/campaign-display/campaign-display.js
kand: With the way it is now, I have to go to another file to figure out what would potentially be passed in here, and that's with prior knowledge of what the structure here is.

If we want to reduce cognitive load, we should be explicit about what goes into this component.

elements/campaign-display/components/campaign-display-root.js
daytonn: Actually, no. I don't think there's any value in manually unpacking the attributes. This is a common pattern for react and I don't see that it's an issue. This is what happens with a wrapper component which just delegates to more specific components.

kand:
:+1:

PR #20, opened 2016-04-11
Campaign display fixes

kand:
Hack for displaying campaign data properly when using separate fields for request v. data.

**Note**: Remove this once we fix the way `store.js` works.


kand:
@collin @daytonn

daytonn:
👍

PR #22, opened 2016-04-12
Consolidate tests

daytonn:
Did some cleanup on the tests so the run without any errors or warnings in the console. They also run a bit faster too. The most likely controversial change is moving tests into a folder of their own in their respective component folders. I did this mainly because I needed to isolate tests that were being noisy and it was much easier once they were consolidated.  I find this organization superior to just putting them in the same place their implementation lives.

Either way, there's a lot of good stuff in here. Let the comments commence!

@kand @collin


.eslintrc
```diff
@@ -189,7 +189,7 @@
     "operator-assignment": [0, "never"],
     "padded-blocks": [0, "never"],
     "quote-props": [0, "as-needed"],
-    "quotes": [1, "single"],
+    "quotes": [1, "avoid-escape"],
```
collin: 👍

lib/bulbs-elements/store/store.js
```diff
@@ -48,11 +48,10 @@ function flattenActions (schema) {
...
   let actions = flattenActions(schema);
-  let wrappedActions = {};
-  iterateObject(actions, (action, actionKey) => {
-    wrappedActions[actionKey] = dispatchFunction.bind(null, store, action);
-  });
-  return wrappedActions;
+  return reduce(actions, (mem, action, key) => {
+    mem[key] = dispatchFunction.bind(null, store, action);
+    return mem;
+  }, {});
```
collin: Can we keep the verbosity with `wrappedActions` and `actionKey` here?

lib/bulbs-elements/store/store.js
```diff
@@ -26,17 +24,19 @@ export default class Store {
...
 }
 
 function flattenActions (schema) {
   let actions = {};
   iterateObject(schema, (field, fieldKey) => {
     iterateObject(field.actions, (action, actionKey) => {
+      if (actions[actionKey]) {
+        invariant(actions[actionKey] === action, `Store: duplicate action named "${actionKey}"`);
+      }
```
collin: Thank you.

lib/bulbs-elements/store/store.js
```diff
@@ -79,20 +74,4 @@ function dispatchFunction (store, action, payload) {
...
-      'color:blue'
-    );
-    console.timeEnd(`dispatch ${action.type}`);
-    console.log('PAYLOAD: ', payload);
-    console.log('PREV VALUE: ', priorFieldValue);
-    console.log('NEXT VALUE: ', nextFieldValue);
-    console.log('PREV STORE: ', priorState);
-    console.log('NEXT STORE: ', nextState);
-    console.groupEnd();
-  }
```
collin: Let's not delete this.

lib/bulbs-elements/store/store.js
daytonn: Sure, I just love an opportunity to use `reduce`

lib/bulbs-elements/store/store.js
```diff
@@ -1,15 +1,13 @@
 import cloneObject from 'clone';
 import iterateObject from 'bulbs-elements/util/iterate-object';
-const DEBUG = false; // placeholder for real implementation
+import invariant from 'invariant';
+import { reduce } from 'lodash';
```
collin: 👍  If we can get a verification that the production build excludes the rest of lodash. Otherwise let's get https://www.npmjs.com/package/lodash.reduce or something.

karma.devicelab.js
```diff
@@ -87,18 +87,7 @@ karmaConfig.customLaunchers = {
...
-  //'bs_android_4_4',
-  //'bs_android_5_0',
-  //'bs_ios_8_3',
-  //'bs_ios_9_1',
-  //'bs_ie_10',
-  //'bs_ie_11',
-  //'bs_safari_8',
   'bs_chrome_47',
-  //'bs_firefox_43',
 ];
```
collin: Please don't delete these.

lib/bulbs-elements/store/store.js
daytonn: I'm not a fan of debug code mixed into production code. Is there a way we can facilitate this more cleanly?

karma.devicelab.js
daytonn: Why are they all commented out? Can we not see this list somewhere in the documentation for device lab?

lib/bulbs-elements/store/store.js
daytonn: Yeah I'm not sure if the form I used will work but I know there's a way to do it with just the methods you need.

karma.devicelab.js
collin: X-browser testing is WIP here.

The correct thing to do here is to not comment these out.

This could live in documentation.

lib/bulbs-elements/store/store.js
collin: Could use a double check, the production build will strip out conditionals that always evaluate to false. If we come up with a way to set DEBUG to true at will in development we can leave this in place.

karma.devicelab.js
daytonn: Does it break if they are uncommented?

lib/bulbs-elements/store/store.js
collin: I'm not sure if this will catch the way DEBUG is done here. But it was working when this was dependent on `process.env.NODE_ENV`

daytonn: Where do we draw the line? Using this convention, how would we discourage logging any function call?

karma.devicelab.js
collin: I think there were runner issues with some environments. 

Travis is using the devicelab karma configuration, so we could use another PR to try them out.

daytonn: Yeah, let's get this working. I generally have a rule that if it's commented out, it gets deleted.

collin:
@daytonn I'm seeing ~4 PR here.

1. lodash
1. moving tests 👍 
1. enabling / deleting devicelab settings
1. removing / finalizing debug logging

Could we get this split out into individual requests so we don't have to consider it as all-or-none?

daytonn:
Yeah, that's a good idea. It kind of started as me trying to get rid of the warnings in the console and I got on a roll. I can definitely split these out. There's also some linter configuration so files like webpack config don't cry.

collin:
Right, everything I didn't call out in that list of four gets a big 👍

daytonn:
This PR has been split up into several others.

PR #23, opened 2016-04-12
Campaign display examples

daytonn:
@kand @collin This adds examples and actually works

### DIFF
--- a/elements/campaign-display/fields/campaign-request-field.js
+++ b/elements/campaign-display/fields/campaign-request-field.js
@@ -19,7 +19,7 @@ const CampaignRequestField = {
 
     fetchCampaignSuccess(state, response, store) {
       state.requestInFlight = false;
-      store.actions.handleFetchComplete(response);
+      setImmediate(() => store.actions.campaign.handleFetchComplete(response));
     },
 
     fetchCampaignFailure(state, response) {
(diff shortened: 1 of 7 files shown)


kand:
👍

elements/campaign-display/fields/campaign-request-field.js
```diff
@@ -19,7 +19,7 @@ const CampaignRequestField = {
 
     fetchCampaignSuccess(state, response, store) {
       state.requestInFlight = false;
-      store.actions.handleFetchComplete(response);
+      setImmediate(() => store.actions.handleFetchComplete(response));
```
collin: How do we feel about a constraint that an action MAY NOT call out to a second action on the same field.

If we can lock down nested actions to only being on other fields, we can avoid any kind of setImmediate/queueing.

In this case, I imagine calling out a second action on the same field would throw an error.

daytonn: This is an action on another field though.

collin: It would require a change to the store, but it wouldn't require any awkward queueing.

collin: So it wouldn't work as-is, but if we disallow nested actions within the same field, we can have a simple change that allows this it happen immediately.

collin: By which I mean the traditional sense of immediately, not the setImmediate nonsense way.

daytonn: I'd like to marinate that a little bit. We want a clean and clear way of describing actions and I'm not convinced actions calling actions, nested or otherwise, is a good idea.

daytonn: What if we didn't merge the actions at all, what happens if we do that?

PR #27, opened 2016-04-13
No console output in tests

daytonn:


karma.devicelab.js
```diff
@@ -87,18 +87,7 @@ karmaConfig.customLaunchers = {
...
-  //'bs_android_4_3',
-  //'bs_android_4_4',
-  //'bs_android_5_0',
-  //'bs_ios_8_3',
-  //'bs_ios_9_1',
-  //'bs_ie_10',
-  //'bs_ie_11',
-  //'bs_safari_8',
   'bs_chrome_47',
-  //'bs_firefox_43',
```
collin: Alright, I think what we actually want to do here, is set up a grid in our `.travis.yml` that passes these browsers in as an environment variable. This way we get better reporting about which device is failing.

daytonn: Can you set that up? I'm not exactly sure what needs to be done.

collin: Yeah, I'll keep that on my plate. 👍  for now.

kand:
👍

PR #28, opened 2016-04-13
Add missing tests

daytonn:


lib/bulbs-elements/test/iterate-object.test.js
```diff
@@ -0,0 +1,20 @@
...
+    iterator = chai.spy(() => {});
+    subject = {
+      one: 1,
+      two: 2,
+    };
+  });
+
+  it('invokes iterator for each key in the object, passing the value of the key, and the key', () => {
+    iterateObject(subject, iterator);
+    console.log(iterator, subject);
```
collin: Left behind?

elements/campaign-display/campaign-display.test.js
```diff
@@ -14,6 +14,11 @@ describe('<campaign-display>', () => {
       campaignUrl,
       display: 'image',
     };
+
+    // Prevent setState warnings spamming the console
+    // We sould investigate if this is an issue with lib/bulbs-elements/store/store.js:60
```
kand: Put a `TODO` here to mark it a bit more clearly as something to investigate later.

kand:
👍

lib/bulbs-elements/test/iterate-object.test.js
daytonn: Oh yeah, I'll get that

elements/campaign-display/campaign-display.test.js
```diff
@@ -14,6 +14,11 @@ describe('<campaign-display>', () => {
...
+    CampaignDisplay.prototype.setState = chai.spy();
```
collin: Somehow we're calling setState before the component is fully mounted in DOM. (React demands that components be mounted before you `setState` on them.)

I think because this test is using shallowRenderer on `<CampaignDisplay/>` it's never mounting in DOM

Maybe store could check the mount status of components before calling setState?

collin:
👍  After comments.

elements/campaign-display/campaign-display.test.js
daytonn: React doesn't like that either, and it's gross.

PR #33, opened 2016-04-19
<campaign-display> <DfpPixel> with render callback passed in as string

kand:

We'd bubble this attribute up to `<campaign-display>` elements in the form of an `on-dpf-pixel-render` attribute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,4 +90,14 @@ Bulbs Elements comes with code generators to help you develop components.
 
 ### Testing
 
+To do a single test run:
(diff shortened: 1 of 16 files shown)


kand:
@collin @daytonn here's the callback passed in as an attribute.

Working on the alternative version where we emit an event for external libraries to use.

kand:
Wrong branch picked to merge into, closing this in favor of https://github.com/theonion/bulbs-elements/pull/34.

PR #32, opened 2016-04-15
Add dfp pixel markup to <campaign-display>

kand:

PR #34, opened 2016-04-19
<campaign-display> <DfpPixel> with render callback passed in as string

kand:

This hook is passed in as a string that references a function on the window.

We'd bubble this attribute up to `<campaign-display>` elements in the form of an `on-dpf-pixel-render` attribute.

### DIFF
--- a/elements/campaign-display/components/dfp-pixel.test.js
+++ b/elements/campaign-display/components/dfp-pixel.test.js
@@ -7,9 +7,18 @@ describe('<campaign-display> <DfpPixel>', () => {
 
   let shallowRenderer = createRenderer();
 
+  context('on render', () => {
+
+    it('should call a callback', () => {
+
+      // TODO : add test code here
+      throw new Error('Not implemented yet.');
+    });
+  });
+
   context('ad unit name', () => {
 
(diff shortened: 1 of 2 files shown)


kand:
@collin @daytonn here's the callback passed in as an attribute.

Working on the alternative version where we emit an event for external libraries to use.

elements/campaign-display/components/dfp-pixel.test.js
```diff
@@ -1,61 +1,94 @@
...
-      chai.spy.on(console, 'error');
-
-      shallowRenderer.render(<DfpPixel placement='junk' />);
 
-      expect(console.error).to.have.been.called.with(
-        'Warning: Failed propType: Required prop `campaignId` was not specified in `DfpPixel`.'
-      );
-    });
-  });
+// TODO : add these back
```
daytonn: Why don't these run?

kand: Just a POC at the moment, I'll add these back in when I merge back into https://github.com/theonion/bulbs-elements/pull/32.

kand:
Closing in favor of https://github.com/theonion/bulbs-elements/pull/36

PR #36, opened 2016-04-19
<campaign-display> <DfpPixel> with render even calling a window configuration

kand:

This takes `window.BULBS_ELEMENTS_ADS_MANAGER` and calls `reloadAds` from that. Requires that the correct `AdsManager` has been set to that variable, and the `AdsManager` interface has a `reloadAds(element)` function.


kand:
@collin @daytonn the third way. I'l go with this one if I get 👍s.

daytonn:
:shipit: 👍

PR #43, opened 2016-04-21
Don't render preamble text if there is no logo or sponsor name

daytonn:
I forgot to check for content before rendering the preamble text

@kand @collin


kand:
👍

collin:
👍

PR #44, opened 2016-04-21
Cropped image

collin:
I had a need for `<CroppedImage/>` in `<bulbs-video>`. So I hoisted it out here.

@kand @daytonn

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,18 @@ Prepare your development environment:
 $ ./scripts/dev-setup
 ```
 
+## Common Components
+
+Some components can be used in more than one element.
+Those components live at `lib/bulbs-elements/components`
+
+### `<CroppedImage>`
+
+*props*
+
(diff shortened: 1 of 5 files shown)


daytonn:
👍

collin:
Paging Doctor Kos to the PR @kand

kand:
Would it be possible to have this as a sibling of other components? I can see this component being used externally as well and I think it's a bit confusing to have components in two places in the project.

collin:
That would take some reorganizing of the project layout. Right now everything in `elements/` is built as an individual js file in `dist/`. Shared code that isn't specific to one element currently lives in `lib/`, Store, register, etc.

If we did create a `<cropped-image>` element I think we should be using the react component version inside of other react components rather than always using the `<cropped-image>`.

```
render () {
  return <CroppedImage/>;
}
```

vs

```
render () {
  return <cropped-image/>;
}
```

kand:
👍

PR #51, opened 2016-04-25
hoisted styles from starwipe

kand:
@collin do we still need this branch?


collin:
Yeah, I need to review this and see if these are the correct styles.

PR #55, opened 2016-04-26
<campaign-display> Add check for active campaign

kand:
Implement awareness of changes in https://github.com/theonion/tunic/pull/50.

### DIFF
--- a/elements/campaign-display/components/campaign-display-root.js
+++ b/elements/campaign-display/components/campaign-display-root.js
@@ -14,19 +14,15 @@ class CampaignDisplayRoot extends Component {
   }
 
   hasImageId() {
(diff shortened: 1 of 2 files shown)


kand:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;h62pPhD7&#x2F;22-for-campaigns-that-are-not-within-a-valid-date-range-tunic-public-campaign-endpoint-should-return-a-200-with-a-running-false-pro"> for campaigns that are not within a valid date range, tunic public campaign endpoint should return a 200, with a running: false property</a>

kand:
@collin @daytonn

collin:
👍

PR #59, opened 2016-04-28
Enzyme

collin:
@kand @daytonn

I was toying with writing `<bulbs-video>` tests with enzyme. I know we wanted to avoid further test fragmentation in this repo, so I checked out if it was sensible to move `<bulbs-poll>` over too.

Was able to quickly replace `AssertJSXEqual` with enzyme (and delete it with extreme prejudice):

```jsx
expect(shallow(<Actual/>).equals(<Expected/>)).to.be.true`.
```

Exposed a few bugs in the tests and needed to bump a few function components up to classes so I could properly test event handlers.

```jsx
function SomeComponent () {
  return <div onClick={this.props.someFunction.bind(null, this.props.someProp)}/>;
}
```

Had to be changed to:
```jsx
class SomeComponent extends React.Component {
  someFunction () {
    this.props.someFunction(this.props.someProp);
  }
  render() { return <div onClick={this.someFunction}/>; }
}
```

So it could be tested:

```jsx
function someFunction () {}
expect(shallow(
  <SomeComponent someFunction={someFunction} someProp="a prop"/>
).equals(
  <div onClick={SomeComponent.prototype.someFunction}/>
)).to.be.true;
```


kand:
👍

daytonn:
👍

PR #63, opened 2016-05-04
Swap out chai spies for sinon

daytonn:
Not that chai spies isn't a perfectly fine library, it's just that sinon is a much more full featured library whose other features we may also want to take advantage of. This is also in a series of recent pull requests that standardize our front-end testing stack.

@kand @collin


kand:
👍

collin:
# As Sayeth Cameron

PR #64, opened 2016-05-06
Fix double sponsor name bug

daytonn:
I noticed that if an image is missing, 2 sponsor names will display. We only want to fallback to the sponsor name if we asked for only an image and we didn't get one.

@kand @collin


spra85:
👍

kand:
👍

PR #65, opened 2016-05-09
Add inner div wrapper to campaign-display

daytonn:
The tests I changed need to be converted to enzyme so the `props.children.props.children` madness is removed but right now I'm just trying not to have to revert the clickhole deploy.
@kand @collin @spra85


kand:
👍

PR #68, opened 2016-05-10
Prevent accessing properties on undefined keys

daytonn:
I noticed after the last patch that I was getting a bunch of errors in the console because I was no longer checking for a valid campaign before attempting to use it's properties.

@kand @collin


kand:
👍

PR #69, opened 2016-05-10
Clean up lint errors and warnings

daytonn:
@kand @collin

### DIFF
--- a/elements/bulbs-poll/components/request-error.js
+++ b/elements/bulbs-poll/components/request-error.js
@@ -1,3 +1,5 @@
+/* eslint-disable react/prop-types */
+
 import React from 'react';
 
 export default function RequestError ({ error, children, reset }) {
--- a/elements/campaign-display/campaign-display.test.js
+++ b/elements/campaign-display/campaign-display.test.js
@@ -1,24 +1,19 @@
 import React from 'react';
 import CampaignDisplay from './campaign-display';
-import { createRenderer } from 'react-addons-test-utils';
+import { shallow } from 'enzyme';
 import fetchMock from 'fetch-mock';
 
 describe('<campaign-display>', () => {
   let subject;
   let placement;
   let props;
-  let shallowRenderer;
   let src;
-  let crop;
   let campaign;
   beforeEach(() => {
-    // TODO: Prevent setState warnings spamming the console
-    // We sould investigate if this is an issue with lib/bulbs-elements/store/store.js:60
-    CampaignDisplay.prototype.setState = sinon.spy();
-
     placement = 'top';
     src = 'http://example.com';
     campaign = {
+      id: 1,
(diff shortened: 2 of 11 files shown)


elements/bulbs-poll/components/request-error.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable react/prop-types */
```
kand: We don't want prop types here?

elements/campaign-display/campaign-display.test.js
```diff
@@ -12,7 +12,7 @@ describe('<campaign-display>', () => {
   let crop;
   let campaign;
   beforeEach(() => {
-    // TODO: Prevent setState warnings spamming the console
+    // Prevent setState warnings spamming the console
```
kand: Leave `TODO` as an obvious marker that we need to come back to this. If we're not planning on that, remove either this comment, the one below, or both.

elements/campaign-display/components/campaign-display-root.js
```diff
@@ -1,4 +1,5 @@
 import React, { PropTypes, Component } from 'react';
+import createFragment from 'react-addons-create-fragment'; // http://facebook.github.io/react/docs/create-fragment.html
```
kand: Remove the comment here, I'd finding this documentation is trivial and first-party to react.

elements/campaign-display/components/campaign-display-root.js
```diff
@@ -131,9 +130,9 @@ CampaignDisplayRoot.propTypes = {
   campaign: PropTypes.shape({
     active: PropTypes.bool,
     clickthrough_url: PropTypes.string,
-    id: PropTypes.number.isRequired,
```
kand: Shouldn't this stay as required?

kand:
Otherwise 👍

elements/campaign-display/components/campaign-display-root.js
daytonn: Essentially nothing is required since we're gracefully handling errors

elements/bulbs-poll/components/request-error.js
daytonn: I'm not entirely sure. We may want them here but there are no tests for this component and I didn't want to investigate it now. I'll add a card to track it (https://trello.com/c/BUEeTOmU/275-remove-ignored-lint-rules-in-bulbs-elements).

elements/campaign-display/components/campaign-display-root.js
daytonn: Actually I agree. I don't think we want to gracefully fail here. If there's something wrong with the CMS that's causing this, sentry will catch it and we want the pain.

PR #73, opened 2016-05-16
new video elements

collin:
@kand @daytonn Left some notes for review.


elements/bulbs-video/bulbs-video.js
```diff
@@ -16,23 +14,57 @@ import 'videojs/dist/video-js/video-js.css';
...
-    if (this.props.src !== props.src) {
+  componentDidUpdate (prevProps) {
+    if (this.props.src !== prevProps.src) {
       this.store.actions.resetController();
-      this.store.actions.fetchVideo(props.src);
+      this.initialDispatch();
+    }
+  }
+
+/*
```
collin: Next task for me related to the video player involves getting redundant requests for the same resource trimmed down. Going to send out an RFC before diving into that in earnest.

elements/bulbs-video/bulbs-video.js
```diff
@@ -47,13 +79,15 @@ Object.assign(BulbsVideo, {
...
+    noEndcard: PropTypes.string,
     src: PropTypes.string.isRequired,
     twitterHandle: PropTypes.isRequired,
   },
 });
 
 registerReactElement('bulbs-video', BulbsVideo);
+
+import './elements/meta';
+import './elements/summary';
```
collin: I'm not sure if it makes sense to move `<bulbs-video-meta>` and `<bulbs-video-summary>` to their own top-level.

Thoughts?

elements/bulbs-video/components/cover.test.js
```diff
@@ -51,7 +51,7 @@ describe('<bulbs-video> <Cover>', function () {
       );
     });
 
-    it('renders a CroppedImage', () => {
+    it('renders an image', () => {
```
collin: As you may know there are problems with using CroppedImage in bulbs-elements.

The local image.js takes over and starts rendering the wrong images.

elements/bulbs-video/elements/meta.js
```diff
@@ -0,0 +1,66 @@
+import React, { PropTypes } from 'react';
```
collin: @daytonn From a few weeks later I'm not sure how vital it is to separate the outermost class from the 'Root' class. Probably not.

elements/bulbs-video/bulbs-video.js
daytonn: We can keep it here until it's obvious.

kand: What's the distinction here between elements and components? Is one a function and the other a class?

elements/bulbs-video/components/cover.test.js
kand: @daytonn weren't we able to use CroppedImage?

elements/bulbs-video/elements/meta.js
daytonn: Not sure what you mean.

elements/bulbs-video/components/revealed.js
```diff
@@ -93,6 +93,10 @@ export default class Revealed extends React.Component {
       twitterHandle: this.props.twitterHandle,
     };
 
+    if (this.props.noEndcard) {
+      delete playerOptions.pluginConfig.endcard;
```
kand: This feels like an anti-pattern. If we don't want an endcard, couldn't we not pass in the endcard config?

elements/bulbs-video/bulbs-video.js
collin: Elements:

* registered as dom elements with `document.registerElement`
* have a special wrapper that marries `document.registerElement` with `ReactDOM.render`

Components
* mere react components (both classes and functions)

elements/bulbs-video/elements/meta.js
collin: I was pushing a hard line on  creating a `Root` component for elements that separates store wiring stuff from top-level rendering logic.

Now that I'm going through and implementing more elements I'm not so attached to that and can't make a cogent argument for it.

elements/bulbs-video/components/revealed.js
collin: These are options coming down through from the videohub JSON api.

kand:
👍

collin:
👍

daytonn:
👍

PR #76, opened 2016-05-16
linter

collin:
@kand @daytonn 

I propose we move linting out of webpack and into `.travis.yml` / `tag-and-release`

At edit-time eslint can be run through your editor rather than in the build script.

### DIFF
--- a/scripts/lint
+++ b/scripts/lint
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 # Summary: lint js files.
 
-node_modules/.bin/eslint elements lib --max-warnings=1 --cache
+npm run lint
--- a/scripts/lint
+++ b/scripts/lint
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 # Summary: lint js files.
 
-npm run lint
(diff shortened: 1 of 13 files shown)


collin:
And we get nicely isolated failure reporting form travis.

scripts/lint
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 # Summary: lint js files.
 
-npm run lint
+node_modules/.bin/eslint elements lib --max-warnings=1
```
collin: must use --max-warnings=1 to get error exit code

scripts/lint
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 # Summary: lint js files.
 
-npm run lint
+node_modules/.bin/eslint elements lib --max-warnings=1 --cache
```
kand: Why not run the npm script here? With this, we have a mix of commands and scripts in `package.json` file.

kand:
I agree with this idea.

daytonn:
That's a good idea. 👍

PR #77, opened 2016-05-16
optimize chunks

collin:
Starting to optimize output based on information from stats.

### DIFF
--- a/scripts/webpack-build-stats
+++ b/scripts/webpack-build-stats
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 export NODE_ENV=production
+set -e
 node_modules/.bin/webpack \
   --bail \
   --color \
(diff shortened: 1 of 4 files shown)


scripts/webpack-build-stats
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+export NODE_ENV=production
```
kand: @daytonn will this export this variable to your shell env once this is run?

kand:
👍

scripts/webpack-build-stats
daytonn: yes `export` sets the variable globally, we can just remove the `export` to make it local. If you want to get nuts you can do `local NODE_ENV=production`.

daytonn:
👍

scripts/webpack-build-stats
collin: This doesn't pollute the parent process.

```
scripts/webpack-build-stats
echo $NODE_ENV 
```
Will not print out `production`

PR #78, opened 2016-05-19
Share tools fixes

collin:
@daytonn @kand Made some changes to the share tools implementation to allow them to play nicer with  analyticsManager.

Also altered the way the outer `<share-tools>` element is registered.

It is now registered as a plain DOM element rather than a React powered element.

This way it's children are simply passed through. When react was powering this element it would blow away the children.

There are some ways we could use React to manage initial in content, something like:

```js
class ReactWrapper extends HtmlElement {
  attachedCallback () {
    let originalChildren = document.createDocumentFragment();
    // pretending that node lists act like arrays
    this.childNodes.forEach((child) => {
      originalChildren.appendChild(child);
    });
    ReactDOM.render(<SomeComponent originalChildren={originalChildren}/>, this.mountPoint);
  }
}
```

Then it would be up to the implementing component to do something with those original children...

```js
class SomeComponent extends React.component {
  render () {
    return <div ref="container"/>;
  }
  componentDidMount () {
    this.props.originalChildren.childNodes.forEach((child) => {
      this.refs.container.appendChild(child);
    });
  }
}
```

Not so great, and I didn't need anything like that here, just wanted to scribble out my thoughts.


kand:
👍

daytonn:
👍

collin:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;dWgzS3Ln&#x2F;11-click-tracking-for-sc-landing-pages"> Click-Tracking for SC Landing Pages</a>

PR #85, opened 2016-06-06
Made host channel targeting a passed in parameter

spra85:
@collin @daytonn @MelizzaP @kand ... Making this a passed in parameter instead of being based on the `global.targeting` value because it will vary based on placement.  This mirrors more closely the old functionality where the player was an iframe embed with a `?target_host_channel=special_coverage_main` query string param


collin:
👍

kand:
👍

PR #86, opened 2016-06-06
Fixing issue where host channel and special coverage slug no longer used for ad targeting

spra85:


spra85:
@collin @MelizzaP @kand @daytonn ... This fixes two issues: 

1) Correctly populating the host channel GA Dimension
2) Correct populating the host channel tag in the VAST url

dimension5 is set to `specialcoverage_main` and that value is also included in the `t` query string param (tags) in the VAST url

http://fix-special-coverage-targeting.test.theonion.com/special/lawn-and-garden/video/3943

kand:
👍

daytonn:
👍

PR #87, opened 2016-06-08
Video carousel refactor

collin:
@kand @daytonn Here's a pass at extracting the carousel stuff into it's own namespace and having a top-level video-carousel wrapper. Works like this:

```html
<bulbs-video-carousel>                                       
  <bulbs-video></bulbs-video>                                
                                                             
  <bulbs-carousel>
    <bulbs-carousel-previous></bulbs-carousel-previous>                           
    <bulbs-carousel-next></bulbs-carousel-next>                           
    <bulbs-carousel-slider>                                  
      <bulbs-carousel-item id='first'></bulbs-carousel-item> 
      <bulbs-carousel-item id='second'></bulbs-carousel-item>
    </bulbs-carousel-slider>                                 
  </bulbs-carousel>                                          
</bulbs-video-carousel>                                      
```


elements/bulbs-carousel/bulbs-carousel.test.js
```diff
@@ -49,22 +64,126 @@ describe('<bulbs-video-carousel-slider>', () => {
...
+      });
+
+      it('returns that item', () => {
+        expect(subject.getActiveCarouselItem()).to.eql(
+          secondItem
+        );
+      });
+    });
+
+    context('multiple items are active', () => {
```
daytonn: What would cause multiple items to be active?

elements/bulbs-carousel/bulbs-carousel.js
```diff
@@ -0,0 +1,174 @@
...
+
+  get carouselItems () {
+    return this.track.children;
+  }
+
+  get firstItem () {
+    return this.carouselItems[0];
+  }
+
+  getActiveCarouselItem () {
```
kand: Can we use es6 getters here as well?

elements/bulbs-carousel/bulbs-carousel.test.js
collin: If for some reason the same video ended up in the recirc list twice.

Mostly paranoia here.

elements/bulbs-carousel/bulbs-carousel.js
```diff
@@ -0,0 +1,174 @@
...
+import { BulbsHTMLElement, registerElement } from 'bulbs-elements/register';
+import { moveChildren } from 'bulbs-elements/util';
+
+import './bulbs-carousel.scss';
+
+export const supportsCalcInTransform = (() => {
+  let div = document.createElement('div');
+  div.style.transform = 'translateX(calc(100% - 10px))';
+  return div.style.transform !== '';
+})();
```
kand: Can this be moved to utilities? Looks like a function that would be useful elsewhere.

elements/bulbs-carousel/bulbs-carousel.js
collin: I  waffled around on these ones.

I'm a fan of getters. I felt that `get caruselItems` and `get firstItem` were obvious cases for getters because they are merely accessing properties. The methods I've prefixed as `getThing` I wasn't so sure about. They are doing potentially more heavy lifting/DOM access.

I'm on the fence, could be pushed either way.

elements/bulbs-carousel/bulbs-carousel.js
collin: Yeah, can do.

elements/bulbs-carousel/bulbs-carousel.js
kand: Okay, I agree with keeping them separate in this case.

elements/bulbs-carousel/bulbs-carousel.js
```diff
@@ -0,0 +1,174 @@
...
+
+export default class BulbsCarousel extends BulbsHTMLElement {
```
kand: The interface of this object seems to be unclearly defined.

Probably most of the time we'd look at the source to figure out what to access, but if we listed out the properties here, wouldn't it show, what are presumably, internal properties--e.g. `slider`, `currentIndex`, `slider`, `handleClick`--mixed in with interface properties--e.g. `carouselItems`, `updateCurrentIndex`, `slideToNext`?

Can some of these properties be hidden via closure or some other strategy? Maybe something to discuss in a larger meeting about how we want to use classes.

collin: I walked down the road of something like:

```js
class BulbsCarousel extends BulbsHTMLElement {
  createdCallback () {
    /* ... */
    this.state = new BulbsCarouselState({
      carouselItems: this.track.children,
      currentIndex: 0,
    });
  }
}
```

`BulbsCarouselState` carried all the `get*` methods, `updateCurrentIndex`, `updateCurrentIndex`, `pageToCarouselItem`, `slideToPrevious`, `slideToNext`, `isOnFirstPage`, and `isOnLastPage`.

`BulbsCarousel` keeps: `createdCallback`, `attachedCallback`, `handleClick`, and `applyState`.

External access to `state` would be discouraged.

collin:
@kand Here's what this would look like with Carousel state extracted into it's own class: https://github.com/theonion/bulbs-elements/pull/91/files

kand:
Is there no good way to have private variables in a `class` in es6? Looked around a bit on the web for anything and it looks like there isn't... weird.

In that case, I guess having `this.state` is fine.

PR #88, opened 2016-06-08
implement new lint rules

collin:
`no-else-return` and `"space-before-function-paren": [1, "never"],`

@daytonn @kand re: some cleanup from review and getting the function paren style tightened down.


collin:
This looks big, but it's just linter stuff.

kand:
We should have spaces before function parens, as with all linter rules we're splitting hairs, but it differentiates function calls from function definitions at a glance, given that function calls do not have spaces between the name and the parens (which obviously is standard).

collin:
I concur with your reasoning for switching `space-before-function-paren` from `never` to `always`.

kand:
👍

daytonn:
👍

PR #89, opened 2016-06-08
return true if campaign is inactive

MichaelButkovic:
@spra85 @kand @daytonn @collin we want to bubble up the displaying of a special coverage to the property cms


spra85:
👍 looks good, just need to resolve conflicts with `master`

collin:
👍

I think you're going to get some linter complaints here for paren spacing, but `npm run lint -- --fix` should be able to auto-correct those ones for you.

kand:
👍

PR #90, opened 2016-06-10
No undefined request

collin:
`<bulbs-video>` was making http requests when the `src` attribute was undefined, leading silly requests like: `GET //example.org/undefined`

This skips making a request if `src` is falsy.

@jmelvnsn @kand @spra85 @daytonn @MelizzaP


spra85:
👍

kand:
👍

PR #92, opened 2016-06-20
Allow muted attribute for bulbs-video element

spra85:
@collin @kand @daytonn @MelizzaP - if you could review, passes through `muted` attribute to the options for the video hub player used here: 

https://github.com/theonion/videohub-player/blob/master/src/player.js#L68


collin:
# 👍  🔇

MelizzaP:
👍

PR #98, opened 2016-06-22
Filtering out the cachebuster param so the analytics are a bit easier to grok

spra85:
@collin @daytonn @kand @MelizzaP ... Will quiet some of the noise with the `aderror` and `adskipped` event logging

### DIFF
--- a/elements/bulbs-video/plugins/google-analytics.test.js
+++ b/elements/bulbs-video/plugins/google-analytics.test.js
@@ -281,6 +281,39 @@ describe('Google Analytics', () => {
     });
   });
 
+  describe('filterQueryString', () => {
+    let googleAnalytics;
+
+    beforeEach(() => {
+      let player = {
+        videoMeta: {
+          channel_name: 'The Onion',
+          player_options: {
+            'shareUrl': 'http://www.theonion.com/r/4053',
+          },
+        },
+        on: sinon.spy(),
+      };
+
+      googleAnalytics = GoogleAnalytics.init(player, 'videoplayer0');
+    });
+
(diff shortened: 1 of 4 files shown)


spra85:
Lot of the lines showing up as changed are because I added `// eslint-disable-line max-len` to make the linter happy

MelizzaP:
this is a 👍 from me, if you could just get that dang linter to agree

elements/bulbs-video/plugins/google-analytics.test.js
```diff
@@ -333,23 +373,57 @@ describe('Google Analytics', () => {
...
+        'aderror: Ad Tag Empty',
+        'http://us-theonion.videoplaza.tv/proxy/distributor/v2?rt=vast_2.0&tt=p&t=1125,the-onion,today-now,main,html5&s=main/the-onion/today-now' // eslint-disable-line max-len
+      );
     });
 
-    it('sends an "aderror" event', () => {
+    it('sends an "aderror" event filtering out rnd value when sandwiched', () => {
+      eventStub = {
+        tag:
+          'http://us-theonion.videoplaza.tv/proxy/distributor/v2?rt=vast_2.0&tt=p&t=1125,the-onion,today-now,main,html5&s=main/the-onion/today-now&rnd=2768879373&foo=bar', // eslint-disable-line max-len
```
kand: For these long lines, I think it would help to break them up so it's more immediately obvious how this is different than/same as the expected output.

spra85: I still don't.  But if others agree with your stance, I will happily change it.

PR #100, opened 2016-06-27
capability to update src/share url on video player

collin:
Lost this in a shuffle.


collin:
I moved this code to another branch manually w/out merging.

PR #102, opened 2016-06-28
Bulbs quiz transplant

daytonn:
I transplanted the quiz code over here. It's not a refactor. I basically moved the jQuery code from one place to another. While this is not the ideal case and we will want to refactor eventually, it's a starting point for cleaning this up.


collin:
Makes sense given most of the source is just copied over.

We need to:

- [ ] add $ to globals for eslint
- [ ] verify $ isn't getting compiled into the dist files

elements/bulbs-quiz/bulbs-quiz-cms.js
```diff
@@ -0,0 +1,17 @@
+import {
+  registerElement,
+  EmbededCMSElement,
+} from 'bulbs-elements/register';
+
+class EmbeddedBulbsQuiz extends EmbededCMSElement {
```
spra85: So minor but we should eventually fix the misspelling of `EmbededCMSElement` to `EmbeddedCMSElement`

elements/bulbs-quiz/bulbs-quiz.scss
```diff
@@ -0,0 +1,3 @@
+bulbs-quiz {
+  display: block;
```
spra85: Wouldn't this be the default?

spra85:
LGTM, should verify the two concerns @collin had

elements/bulbs-quiz/bulbs-quiz.test.js
```diff
@@ -0,0 +1,67 @@
...
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  describe('TestQuiz', () => {
+    it('instantiates a TestQuiz when the quiz style is test', (done) => {
+      fixtureContainer.appendChild(subject);
+      setImmediate(() => {
```
kand: What is `setImmediate` used for in these tests?

kand:
👍

elements/bulbs-quiz/bulbs-quiz.scss
daytonn: I don't think it is in the case of custom elements. @collin can you confirm this?

PR #104, opened 2016-07-06
Transplant slideshow and small refactor

daytonn:


daytonn:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;CSmrItc4&#x2F;18-create-a-slideshow-component"> Create a Slideshow component</a>

bower.json
```diff
@@ -17,7 +17,8 @@
     "tests"
   ],
   "devDependencies": {
-    "jquery": "^2.2.3"
+    "jquery": "^2.2.3",
+    "flexslider": "^2.6.1"
```
kand: Is this a new dependency? Or one that wasn't listed before?

karma.base.js
```diff
@@ -21,6 +21,7 @@ module.exports = {
     // and sets up test helpers (chai-enzyme, etc.)
     path.join(__dirname, 'node_modules/whatwg-fetch/fetch.js'),
     path.join(__dirname, 'test/index.js'),
+    'test/fixtures/**/*',
```
kand: If we don't need `path.join` in those other paths, we should remove them there.

package.json
```diff
@@ -56,6 +56,7 @@
     "extract-text-webpack-plugin": "^1.0.1",
     "fetch-mock": "^4.3.1",
     "file-loader": "^0.8.5",
+    "flexslider": "^2.6.0",
```
kand: If we have a `flexslider` dep here, can we remove the bower one?

kand:
Otherwise 👍

collin:
👍

bower.json
daytonn: This is a jQuery plugin that exists in the clickhole site. We may want it explicitly in this code but for now it's just something I need in the tests.

karma.base.js
daytonn: Yeah, I don't think `path.join` is necessary, I just opted out of using it because `**/*` is a glob and not actually a path so it seemed weird to use it.

package.json
daytonn: No, karma is using bower to load this. This is a shit jQuery plugin that has no module system support.

PR #112, opened 2016-07-15
Fix start percentage reference

daytonn:
This fixes split pics from starting at the default percentage when a custom percentage was given.


MelizzaP:
👍

kand:
👍

PR #115, opened 2016-07-18
campaign in rail player is name-only

collin:
@MelizzaP @spra85 @daytonn @kand @shawncook @jmelvnsn 

Overlooked this in the rail-player.


spra85:
👍

PR #116, opened 2016-07-19
update rail player targeting

collin:
@spra85 Like this?


spra85:
@collin, looks right 👍 ... I'm assuming that will work to set the correct prop within the `Revealed` component.

collin:
Yeah, it looks like the correct one to me: https://github.com/theonion/bulbs-elements/blob/master/elements/bulbs-video/components/revealed.js#L72

collin:
@spra85 Will we be able to confirm this easily on a test deploy?

spra85:
Yeah, I can check it for you really quick

PR #117, opened 2016-07-20
Make styles available to carousel play button

jmelvnsn:
@spra85 @collin Fixes unstyled share buttons on carousel items


jmelvnsn:
Actually nevermind. I want to move this to something more global.

jmelvnsn:
@spra85 @collin This is better

spra85:
Seems fine, should just make sure the rail-player with branded content still has the same styling it does now, it has a darker fill, etc.

PR #120, opened 2016-07-25
load on demand

collin:
This implements load-on-demand per component. Like betty-cropper's image.js.

In this example, `ComponentToLoadOnDemand` would not be rendered until it was within 200px of the viewport.

```js
import { LoadOnDemand } from 'bulbs-elements/util';

class ComponentToLoadOnDemand extends React.Component {
  render () {
    return <h1> Hoopla </h1>
  }
}

export default function WrappedComponent (props) {
  return <LoadOnDemand
    component={ComponentToLoadOnDemand}
    componentProps={props}
  />;
}
```

Concretely, we could us this with the `<rail-player>` element:

In `..../rail-player.js`:

```js
export class RailPlayer extends BulbsElement {
  /* Rail player implementation */
}

registerReactElement('rail-player', (props) => {
  return <LoadOnDemand
    component={RailPlayer}
    componentProps={props}
  />;
});
```


collin:
@daytonn @kand Starting to explore betty-croppy style load on demand to bulbs-elements.

Thoughts? Does this seem like a sane way to compose this kind of higher order component?

lib/bulbs-elements/util/load-on-demand.js
```diff
@@ -0,0 +1,86 @@
...
+function maybeLoadComponent (component) {
+  if (shouldComponentLoad(component)) {
+    loadComponent(component);
+    dequeue(component);
+  }
+}
+
+function shouldComponentLoad (component) {
+  let rect = component.refs.sentinel.getBoundingClientRect();
+  return rect.bottom > -200 && rect.top < window.innerHeight + 200;
```
spra85: I think we want this to be a configuration of sorts

kand:
This looks sweet 💯

lib/bulbs-elements/util/load-on-demand.js
```diff
@@ -0,0 +1,86 @@
...
+    dequeue(this);
+  }
+
+  render () {
+    let WrappedComponent = this.props.component;
+    if (this.state.load) {
+      return <WrappedComponent {...this.props.componentProps}/>;
+    }
+
+    return <div ref="sentinel" {...this.props}/>;
```
spra85: What's `ref="sentinel"` in this case?

spra85:
I was envisioning using an attribute instead of an additional wrapped component layer, something like: 

`<rail-player src="/video/12457.json" lazy>`

I just feel like there might be scenarios where you want to use the component but you know they're at the top of the page and you might not want to wait for the `resize` or `scroll` events to fire so baking in the flexibility to decide lazy or not as an attr seems to make sense.  Thoughts?

lib/bulbs-elements/util/load-on-demand.js
collin: That makes sense.

lib/bulbs-elements/util/load-on-demand.js
collin: That gives us a hook to the `<div>` element `component.refs.sentinel`. That's the element being used to check for position within the viewport.

collin:
@spra85 As implemented, this will do an immediate check on first-render without waiting for `scroll` or `resize` events.

collin:
It could still be switched on/off with an attribute.

spra85:
Ahh, that makes sense.  I'm a fan.  Nice start

elements/lazy-iframe/lazy-iframe.js
```diff
@@ -0,0 +1,16 @@
...
+  loadOnDemand((props) => (
+    <iframe
+      {...props}
+      style={{
+        width: '100%',
+        height: '100%',
+      }}
+    />
+  ))
+);
```
collin: @daytonn @kand @spra85 

`<lazy-iframe>` will be first used in the 1996-week coverage to avoid loading all the youtube embed iframes until they are closer to the viewport.

collin: Looking to merge this.

This strategy could also be used with the `bundle-loader` to lazily load implementation for elements on demand: https://gist.github.com/collin/9b0a6ae2093ed70ad4d46b94a3fc3c3b

collin:
Alright, I'm taking

> :100:

and 

> I'm a fan

as approval.

The outstanding comment was distance from viewport configurability. I added a comment with implementation guides for how to make the distance from the bottom/top configurable for if/when we need that.

PR #122, opened 2016-07-27
Medal tracker styles

shawncook:
Hey @daytonn, I added a few flexbox styles that should take care of the vertical alignment of the medal numbers and contender names. These were styles I had written to override it in the clickhole project, and they definitely belong here instead.


daytonn:
👍

collin:
👍

PR #124, opened 2016-07-29
set location.hash if comments are on current page

collin:
@daytonn @jmelvnsn @shawncook @spra85 

If `<share-via-disqus>` is on the same page the comments are, we don't need to open a new window, just jump to the comments by setting `window.location.hash =  '#comments'`;


collin:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;avgHtMfL&#x2F;19-add-share-tools-to-feature-d-t-m"> Add share tools to feature (D&#x2F;T&#x2F;M)</a>

PR #125, opened 2016-08-01
Notifications

benghaziboy:
@collin please shame me


elements/notification-container/notification-container.js
```diff
@@ -0,0 +1,86 @@
...
+
+    if (this.state.local_notification_ids.length > 0) {
+      notifications = this.removeLocalNotifications(notifications);
+    }
+
+    if (notifications.length > 0) {
+      this.setState({ notification: notifications[0] });
+      this.setState({
+        local_notification_ids: this.state.local_notification_ids.concat([notifications[0].id])
+      });;
```
collin: These can be squashed into one `setState`

```js
setState({
  notification: notifications[0],
  local_notification_ids: this.state.local_notification_ids.concat([notifications[0].id]),
});
```

elements/notification-container/notification-container.js
```diff
@@ -0,0 +1,86 @@
...
+    } else if (this.state.next !== null) {
+      this.requestNotifications(this.state.next);
+    }
```
collin: It doesn't look like next is used outside this function.

I think you could avoid setting it as state and 

```js
else if (response.next !== null) {
  this.requestNotifications(response.next);
}
```

elements/notification-container/notification-container.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  removeLocalNotifications (notifications) {
+    for (var i = notifications.length - 1; i >= 0; i--) {
+      if (this.state.local_notification_ids.indexOf(notifications[i].id) != -1) {
+        notifications.splice(i, 1);
+      }
+    }
+    return notifications;
+  }
```
collin: We use `foreach` and then for style, here I would create a new array, and build it up rather than mutate the passed-in array.

```js
var withoutLocal = [];
notifications.forEach((notification) => {
  if (/* conditional */) {
    withoutLocal.push(notification);
  }
});
return withoutLocal;
```

collin:
@benghaziboy: http://eslint.org/docs/user-guide/integrations

Also you're going to get a lot of grief from the linter.

collin:
I'm also not sure how to feel about the mixin thing.

FB is tracking advising against the mixin api with an eye towards deprecating it as soon as they can.

I think we can get what we need from localStorage with a couple small changes:

Loading from localStorage in the constructor:
```js
constructor () {
  /* super, etc */
  var idsFromStorage = JSON.parse(
    localStorage.getItem('onion-notifications-local-ids') || '[]'
  );
  this.state = {
    local_notification_ids: idsFromStorage,
  };
}
```

And writing to localStorage when setting sate:
```js
var idsForState = this.state.local_notification_ids.concat([notifications[0].id]);
this.setState({
  local_notification_ids: idsForState,
});
localStorage.setItem(
  'onion-notifications-local-ids',
  JSON.stringify(idsForState)
);
```

benghaziboy:
@collin and again

collin:
Yeah, I think this is better using localStorage directly.

I think we could still reduce the calls to `setState`, but I'd say move on with the feature and maybe review a little more after the release.

elements/notification-container/notification-display.js
```diff
@@ -0,0 +1,45 @@
+import BulbsElement from 'bulbs-elements/bulbs-element';
+import invariant from 'invariant';
+import React, { PropTypes } from 'react';
+import { registerReactElement } from 'bulbs-elements/register';
+
+class NotificationDisplay extends BulbsElement {
```
collin: This sub-component should `extends React.Component`

elements/notification-container/notification-display.js
```diff
@@ -0,0 +1,45 @@
...
+}
+
+Object.assign(NotificationDisplay, {
+  displayName: 'NotificationDisplay',
+  propTypes: {
+    notification: PropTypes.object.isRequired,
+  },
+});
+
+registerReactElement('notification-display', NotificationDisplay);
```
collin: And it should not be registered. Only the outer-most element is registered.

elements/notification-container/notification-container.js
```diff
@@ -0,0 +1,96 @@
...
+
+  constructor (props) {
+    invariant(!!props.src, 'notification-container component requires a src');
+    super(props);
+    let idsFromStorage = JSON.parse(
+      global.localStorage.getItem('onion-local-notification-ids') || '[]'
+    );
+    this.state = {
+      next: null,
+      notifications_are_off: JSON.parse(
```
collin: Looks like this state is not used. I see the `notificationsAreActive` method does this check instead. Should remove from state.

elements/notification-container/notification-container.js
```diff
@@ -0,0 +1,96 @@
...
+        window.sessionStorage.getItem('onion-notifications-off') || 'false'
+      ),
+      local_notification_ids: idsFromStorage,
```
collin: I agree with the readability of the snake case, but according to conventions this is `localNotificationIds` :(

elements/notification-container/notification-container.js
```diff
@@ -0,0 +1,96 @@
...
+      .then(this.handleRequestSuccess.bind(this));
+  }
+
+  componentDidMount () {
+    this.requestNotifications(this.props.src);
+  }
+
+  handleRequestSuccess (response) {
+    let notifications = response.results;
+    this.setState({ next: response.next });
```
collin: The `next` state isn't used in any other functions, no need to set it as state.

See comment @ line 56

elements/notification-container/notification-container.js
```diff
@@ -0,0 +1,96 @@
...
+
+    if (this.state.local_notification_ids.length > 0) {
+      notifications = this.removeLocalNotifications(notifications);
+    }
+
+    if (notifications.length > 0) {
+      this.setState({ notification: notifications[0] });
+      this.setState({
+        local_notification_ids: this.state.local_notification_ids.concat([notifications[0].id]),
+      });
```
collin: Can be a single call to `setState`

```
this.setState({
  local_notification_ids: /* ... */,
  notifications: /* ... */,
});
```

elements/notification-container/notification-container.js
```diff
@@ -0,0 +1,96 @@
...
+      global.localStorage.setItem(
+        'onion-local-notification-ids', JSON.stringify(this.state.local_notification_ids)
+      );
+    }
+    else if (this.state.next !== null) {
```
collin: Could just check `if (response.next)`

package.json
```diff
@@ -98,8 +98,10 @@
     "react-addons-create-fragment": "^15.0.2",
     "react-addons-test-utils": "^15.0.0",
     "react-dom": "^15.0.0",
+    "react-localstorage": "^0.3.0",
     "react-proptypes-url-validator": "0.0.1",
     "react-router": "^1.0.3",
+    "react-mixin": "2.0.0",
```
collin: Don't forget to remove these deps.

collin:
👍 Cool beans.

Hey let's hear it for the beans.

PR #138, opened 2016-08-12
Enabling VTT Captions and making them the default for the rail player

spra85:
@collin @daytonn @kand @MelizzaP.  For Monday review, this PR will enable closed captioning on our video player as an option if it finds a `text/vtt` closed caption track in the list of sources.  It will also make it so the default option of whether to display them is enabled for the rail-player.  Also, docs for reference to JWPlayer are here (https://support.jwplayer.com/customer/portal/articles/1407438-adding-closed-captions)

Test link with an example here: 
http://closed-captioning.test.onionstudios.com/videos/watch-coach-speedman-make-a-difference-in-the-life-of-a-young-middle-manager-4257

You'll need to enable via the `CC` hover in the video player controls

### DIFF
--- a/elements/bulbs-video/bulbs-video-examples.js
+++ b/elements/bulbs-video/bulbs-video-examples.js
@@ -72,6 +72,20 @@ let examples = {
         `;
       },
     },
+    'VTT Captioning Example': {
+      render () {
+        return `
+          <bulbs-video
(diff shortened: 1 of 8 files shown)


kand:
Once the travis reported lint errors are fixed, 👍

spra85:
Should be fixed with most recent commit @kand

daytonn:
👍

PR #141, opened 2016-08-17
a <video is="bulbs-cinemagraph"> is born

collin:
Usage: This custom element uses the element extension form of custom elements.

So far we've been using fully custom elements.

Stuff like:
```html
 <cool-custom-element></cool-custom-element>
```
and
```js
document.createElement('cool-custom-element');
```

But you can also extend built-in elements, like this:
```html
 <video is="cool-video-extensions"></video>
```
and
```js
document.createElement('video', 'cool-video-exensions');
```
======

Here is how we can do cinemagraphs (that play inline on ios!!!)

with src attr:
```html
<video src="/video.mp4" is="bulbs-cinemagraph" cinemagraph-duration="1.55"/>
```

with source element:
```html
<video is="bulbs-cinemagraph" cinemagraph-duration="1.55">
  <source src="/video.mp4"/>
</video>
```
Duration needs to be figured out and set by hand to prevent the ios polyfill from showing a frame of black at the loop seam.

@shawncook @jmelvnsn @MelizzaP 

Let's get cinemagraphic!

### DIFF
--- a/elements/bulbs-cinemagraph/bulbs-cinemagraph.test.js
+++ b/elements/bulbs-cinemagraph/bulbs-cinemagraph.test.js
@@ -1,19 +1,25 @@
-import * as iphoneInlineVideo from 'iphone-inline-video';
+import { InViewMonitor } from 'bulbs-elements/util';
 import './bulbs-cinemagraph';
+import * as iphoneInlineVideo from 'iphone-inline-video';
 
 describe('<video is="bulbs-cinemagraph">', () => {
   let subject;
 
   beforeEach(() => {
     subject = document.createElement('video', 'bulbs-cinemagraph');
+    sinon.spy(InViewMonitor, 'add');
+    sinon.spy(InViewMonitor, 'remove');
+    sinon.spy(subject, 'pause');
+    sinon.spy(subject, 'play');
+  });
+
+  afterEach(() => {
+    InViewMonitor.add.restore();
+    InViewMonitor.remove.restore();
   });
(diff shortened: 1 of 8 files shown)


collin:
@kand @daytonn

elements/bulbs-cinemagraph/bulbs-cinemagraph.test.js
```diff
@@ -0,0 +1,30 @@
...
+    expect(subject.getAttribute('webkit-playsinline')).to.eql('true');
+  });
+
+  it('overrides the video duration', () => {
+    subject.setAttribute('cinemagraph-duration', '4.55');
+    expect(subject.duration).to.eql(4.55);
+  });
+
+  describe('attachedCallback', () => {
+    // Not sure how to assert the makeVideoPlayableInline function was called.
```
kand: Is this something that needs to be filled in?

kand:
👍

elements/bulbs-cinemagraph/bulbs-cinemagraph.test.js
collin: In theory it would be nice to test that the function is called, but I can't figure out a way to do it.

collin:
Only Chrome is failing.

But that's because the testrunner isn't running (even on retry, assuming job is borked).

Confirmed passing locally.

collin:
I confirmed the IE test manually in browserstack.

PR #143, opened 2016-08-18
a utility that monitors when items enter/exit view

collin:
This utility allows an element to opt into two new events:

`enterviewport`

`exitviewport`

It can be handily used in the lifecycle hooks of an element.

```js
import { InViewMonitor } from 'bulbs-elements/status';

class RealCoolElement extends BulbsHTMLElement {
  attachedCallback () {
    InViewMonitor.add(this);
  }

  detachedCallback () {
    InViewMonitor.remove(this);
  }
}

registerElement('real-cool-element', RealCoolElement);

let coolElement = document.create('real-cool-element');

coolElement.addEventListener('enterviewport', function () {});
coolElement.addEventListener('exitviewport', function () {});
```

This will be used in the `<video is="bulbs-cinemagraph">` to play/pause it when it enters/leaves the viewport.

@kand @daytonn

### DIFF
--- a/lib/bulbs-elements/util/in-view-monitor.js
+++ b/lib/bulbs-elements/util/in-view-monitor.js
@@ -1,0 +1,65 @@
+const isElementInViewport = (el) => {
+  let rect = el.getBoundingClientRect();
+
+  return rect.bottom > 0 &&
+    rect.right > 0 &&
+    rect.left < (window.innerWidth || document.documentElement.clientWidth) &&
+    rect.top < (window.innerHeight || document.documentElement.clientHeight)
+  ;
+};
+
+let inViewId = 0;
+let initialized = false;
+let animationRequest;
+let all = {};
+
+const init = () => {
+  if (initialized === false) {
+    initialized = true;
+    window.addEventListener('scroll', handleDisplayMutation);
+    window.addEventListener('resize', handleDisplayMutation);
+  }
+};
+
+const maybeEmitEvent = (one) => {
+  let inView = isElementInViewport(one.element);
+  if (inView && one.state === 'init' || one.state === 'out') {
+    one.state = 'in';
+    let event = new Event('enterviewport');
+    one.element.dispatchEvent(event);
+  }
+  else if (!inView && one.state === 'in') {
+    one.state = 'out';
+    let event = new Event('exitviewport');
+    one.element.dispatchEvent(event);
+  }
+};
+
+const handleDisplayMutation = () => {
+  if (!animationRequest) {
+    animationRequest = requestAnimationFrame(() => {
+      animationRequest = null;
+      Object.keys(all).forEach((key) => {
+        maybeEmitEvent(all[key]);
+      });
+    });
+  }
+};
+
+export default {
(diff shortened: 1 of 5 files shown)


lib/bulbs-elements/util/in-view-monitor.js
```diff
@@ -0,0 +1,65 @@
+const isElementInViewport = (el) => {
+  let rect = el.getBoundingClientRect();
+
+  return rect.bottom > 0 &&
```
collin: I think this check is wrong.

kand:
Can we add a top/bottom distance from top/bottom of screen for potential reuse with the reading list?

collin:
@kand like this?

```js
class RealCoolElement extends BulbsHTMLElement {
  attachedCallback () {
    InViewMonitor.add(this, {
      distanceFromTop: -200
      distanceFromBottom: 200
    });
  }

  detachedCallback () {
    InViewMonitor.remove(this);
  }
}
```

Which I think would also get us calculations based on header/footer heights.

```js
class RealCoolElement extends BulbsHTMLElement {
  attachedCallback () {
    InViewMonitor.add(this, {
      get distanceFromTop () {
        return /* calculation based on header height */;
      }
    });
  }
}
```

kand:
Something like that yea, although we might want to be careful how we pass around header height. I remember some issues around header height because it changes from static to fixed and the height changes.

Could these parameters be functions that are evaluated and return the amount to measure by?

collin:
Yes, defining the option as a getter would let us use just-in-time calculations:

```js
InViewMonitor.add(this, {
  get distanceFromTop () {
    return document.querySelector('#site-header').offsetHeight;
  }
});
```

lib/bulbs-elements/util/in-view-monitor.js
```diff
@@ -0,0 +1,65 @@
...
+const init = () => {
+  if (initialized === false) {
+    initialized = true;
+    window.addEventListener('scroll', handleDisplayMutation);
+    window.addEventListener('resize', handleDisplayMutation);
+  }
+};
+
+const maybeEmitEvent = (one) => {
+  let inView = isElementInViewport(one.element);
```
daytonn: what is `one` here?

daytonn: maybe `all` would be better named `components` and `one` is better named `component`

collin: Yeah, the `all` collection stashes items that look like:
```js
{
  element: domNode,
  state: 'init' || 'in' || 'out',
}
```

I'm all-in on a more descriptive name, but I'm not sure what would be best here.

collin:
@kand This now takes options `distanceFromTop` and `distanceFromBottom`

Both getter and number values work:

getter:
```js
InViewMonitor.add(this, {
  get distanceFromTop () {
    return document.querySelector('#site-header').offsetHeight;
  },
});
```

value
```js
InViewMonitor.add(this, {
  distanceFromTop: document.querySelector('#site-header').offsetHeight,
});
```

PR #145, opened 2016-08-24
This test randomly flickers, the feature is expired so I'm removing t…

daytonn:


collin:
👍

collin:
@daytonn any objection to merging this now?

PR #148, opened 2016-09-06
Fix commons size

collin:
Manually collect common chunks.

@daytonn @spra85 @daytonn @mparent61 

We had a major regression in our webpack builds.

I thought I had verified the commons chunk plugin as working. But currently all shared scripts are being compiled into every element.

`bulbs-poll`, `bulbs-video`, `share-tools`, etc. all include common assets, this is ballooning each script by quite a lot.

This also appears to be the cause of development environments crashing on page load. This is not the root cause, but serving so many large files in development seems to trigger it. In development mode this adds ~800K to each bulbs-element (sourcemaps, etc.).

### DIFF
--- a/webpack.base.js
+++ b/webpack.base.js
@@ -48,10 +48,10 @@ entries['dist/vendor'] = [
   './lib/bulbs-elements/register',
   './lib/bulbs-elements/store',
   './lib/bulbs-elements/bulbs-element',
-	'./lib/bulbs-elements/util/load-on-demand',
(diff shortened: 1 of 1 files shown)


collin:
Merging for now because this is adding weight to all page loads and cameron is about to deploy a patch to a property. Can discuss particulars and make further changes as we go.

mparent61:
Would this have increased `collectstatic` build output as well?

PR #150, opened 2016-09-07
move onion-notifications-off to local instead of session storage

benghaziboy:


collin:
This seems like a rather blunt instrument. What's the purpose for it?

benghaziboy:
the purpose of the feature or the purpose of the change?

collin:
Of the change. It seems like this change would also alter (greatly restrict) the uses of the notifications feature.

benghaziboy:
sessionStorage doesn't persist across tabs, localStorage does. Needs to change.

benghaziboy:
The idea is if someone closes out of their notification, they should not see another until the browser is closed.

collin:
Right, but `localStorage` won't clear out when the browser is closed. This flag would live forever.

Might have to do something like set a timestamp and do something like:

```
notifications_are_active = 'onion-notifications-off' > n days old
```

benghaziboy:
@collin

benghaziboy:
@kand

PR #151, opened 2016-09-12
Dockerize tests

mparent61:
Move all tests into Docker (prep for move to MonoRepo).

Build matrix has been eliminated (can't use it with MonoRepo). Instead single script triggers all browser flavor tests. Overall builds now take 7-8 minutes (vs 2-3 minutes for each of the 5 build matrix builds). 

One downside is it's harder to see which browser(s) failed b/c no longer have nice Build Matrix Travis UI (instead need to dig into Travis logs).


Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+ENV PATH /project/node_modules/.bin:$PATH
+
+# Only build Bower on config change
+ADD bower.json /project/
+RUN bower install --config.interactive=false install --allow-root --force-latest
+
+ADD . /project
+
+# TODO: webpack build? (so don't need to run for every test)
```
collin: We're running webpack through karma, not sure how/if we could cache that.

mparent61: OK won't worry about caching for now

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+RUN apt-get update \
+    && apt-get upgrade -y \
+    && rm -rf /var/lib/apt/lists/*
+
+# TODO: is gcc4.9 (default) OK?
+#RUN apt-get install -y software-properties-common
+## TODO: Merge with above
+#RUN add-apt-repository ppa:ubuntu-toolchain-r/test
+#RUN apt-get install -y g++-4.8
```
mparent61: @collin Do we need g++4.8? By default 4.9 is installed, and tests pass...

Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM node:5.12
```
mparent61: @collin: Should I stick with 5.9, or OK to update to latest 5.12?

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+ENV PATH /bulbs/node_modules/.bin:$PATH
+
+# Only build Bower on config change
+ADD bower.json /bulbs/
+RUN bower install --config.interactive=false install --allow-root --force-latest
+
+ADD . /bulbs
+
+# TODO: Move to top
+# RUN apt-get install xvfb
```
mparent61: @collin Don't think we need xvfb anymore b/c of Browsertack?

scripts/ci
```diff
@@ -1,2 +1,19 @@
 #!/usr/bin/env bash
-./node_modules/.bin/karma start karma.devicelab.js --single-run
+
+# mparent(2016-09-12): Temp Travis/Docker Testing
+export CHROME_BIN=chromium-browser
+export DISPLAY=:99.0
+sh -e /etc/init.d/xvfb start
```
mparent61: ditto

Dockerfile
collin: We're good to take those things out.

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+# Only build NPM on config change
+ADD package.json /bulbs/
+RUN npm --loglevel=warn install
+
+ENV PATH /bulbs/node_modules/.bin:$PATH
+
+# Only build Bower on config change
+ADD bower.json /bulbs/
+RUN bower install --config.interactive=false install --allow-root --force-latest
```
mparent61: @collin: To `--force-latest` or not to `--force-latest`?

Dockerfile
collin: So, according to Node versioning documentation, we shouldn't be using the 5.x series.

https://github.com/nodejs/LTS#lts_schedule

Looks like 6.x is going to be the next LTS very soon, so I think we should try switching to that.

Given our test coverage I'd be comfortable pining this as loosely as this `FROM node:6`

mparent61: 6 it is, thx

Dockerfile
collin: I think no. Could be dangerous. Better for us to be notified here

Dockerfile
collin: Maybe this was added for coverage, or linting? But if it passes now without it, let's strip it out.

scripts/ci
```diff
@@ -1,2 +1,14 @@
...
+BROWSERS=(
+    bs_chrome_50
+    bs_firefox_47
+    bs_safari_8
+    bs_ie_11
+)
+
+for B in ${BROWSERS[@]}; do
+    BROWSER=$B ./node_modules/.bin/karma start karma.devicelab.js --single-run
+done
```
collin: Let's be sure to run the linter too.

`scripts/lint`

mparent61:
@collin Ready for review.

scripts/ci
mparent61: Left this command in `.travis.yml`, unless you'd rather move it here.

collin: Ah, is that going to change when we go to omni?

mparent61: Let's leave it for now, and we'll update it when it goes into OMNI (slightly different layout)

Dockerfile
mparent61: Removed

scripts/ci
mparent61: Changed my mind, moved Lint into this script

collin:
👍

mparent61:
@kand @daytonn   Passed @collin's gauntlet, ready for additional review.

kand:
Sweet 👍

collin:
@mparent61 hold up a second, I have a change to further clean up the output.

collin:
We were using the `progress` reporter that would print a line per test. This one will only print the failures.

mparent61:
Nice

collin:
Still have a bunch of warnings etc.

But it's looking good.

PR #154, opened 2016-09-14
campaign-product-shot & campaign-brand-messaging

MichaelButkovic:


MichaelButkovic:
@collin @kand @daytonn @MelizzaP @spra85 Ready for review

elements/campaign-brand-messaging/campaign-brand-messaging.js
```diff
@@ -0,0 +1,50 @@
...
+    invariant(!!this.hasAttribute('src'), 'campaign-brand-messaging component requires a src');
+    fetch(this.getAttribute('src'))
+      .then(filterBadResponse)
+      .then(getResponseJSON)
+      .then(this.handleRequestSuccess.bind(this))
+      .catch(this.handleRequestError);
+  }
+
+  detachedCallback () {
+    console.log('Detached campaign-brand-messaging');
```
spra85: We should omit the callback if we're only logging something

elements/campaign-brand-messaging/campaign-brand-messaging.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  attributeChangedCallback (name, previousValue, value) {
+    if (document.body.contains(this)) {
+      if (name === 'src' && previousValue !== value) {
+        this.attachedCallback();
+      }
+
+      console.log(
+        'Attribute Changed campaign-brand-messaging changed ${name} from: ',
```
spra85: Do we want to have this kind of noisy logging in the production version of this?

elements/campaign-product-shot/campaign-product-shot.js
```diff
@@ -0,0 +1,50 @@
...
+} from 'bulbs-elements/register';
+import { filterBadResponse, getResponseJSON } from 'bulbs-elements/util';
+import invariant from 'invariant';
+
+class CampaignProductShot extends BulbsHTMLElement {
+
+  handleRequestSuccess (data) {
+    this.innerHTML =
+      `<div class='campaign-product-shot'>
+        <img src='${data.product_shot_url}'>
```
spra85: We should check to make sure the `product_shot_url` is not empty prior to setting it on the img src.  Possibly log an error if so

elements/campaign-brand-messaging/campaign-brand-messaging.js
```diff
@@ -0,0 +1,46 @@
...
+        <span>${data.brand_messaging}</span>
+      </div>`;
+  }
+
+  handleRequestError (error) {
+    console.log(error);
+  }
+
+  attachedCallback () {
+    invariant(!!this.hasAttribute('src'), 'campaign-brand-messaging component requires a src');
```
daytonn: can this just be `this.hasAttribute('src')`? Doesn't it already return a boolean?

elements/campaign-brand-messaging/campaign-brand-messaging.js
```diff
@@ -0,0 +1,46 @@
...
+  BulbsHTMLElement,
+} from 'bulbs-elements/register';
+import { filterBadResponse, getResponseJSON } from 'bulbs-elements/util';
+import invariant from 'invariant';
+
+class CampaignProductShot extends BulbsHTMLElement {
+
+  handleRequestSuccess (data) {
+    if (!data.brand_messaging) {
+      console.log('Brand messaging not set on this campaign');
```
collin: If we're going to do messages like this, we should also log the specific element:

```
console.log('Brand messaging not set on this campaign', this);
```

So we can right click/reveal in elements panel.

elements/campaign-product-shot/campaign-product-shot.js
```diff
@@ -0,0 +1,46 @@
...
+      </div>`;
+  }
+
+  handleRequestError (error) {
+    console.log(error);
+  }
+
+  attachedCallback () {
+    invariant(!!this.hasAttribute('src'), 'campaign-product-shot component requires a src');
```
daytonn: Same here, I don't the the double negation is necessary.

elements/campaign-brand-messaging/campaign-brand-messaging.js
```diff
@@ -0,0 +1,46 @@
...
+      if (name === 'src' && previousValue !== value) {
+        this.attachedCallback();
+      }
+    }
+  }
+}
+
+registerElement('campaign-brand-messaging', CampaignProductShot);
+
+export default CampaignProductShot;
```
collin: I think this is an issue of the generator, but I don't see any reason not to inline the `export default` with the class definition:

```
export default class CampaignProductShot extends BulbsHTMLElement
```

If you change this, we should also update the generator: https://github.com/theonion/bulbs-elements/blob/master/generator-bulbs-elements/generators/dom-element/templates/%7B%7B%20elementPathName%20%7D%7D.js#L7

elements/campaign-product-shot/campaign-product-shot.js
```diff
@@ -0,0 +1,46 @@
...
+      if (name === 'src' && previousValue !== value) {
+        this.attachedCallback();
+      }
+    }
+  }
+}
+
+registerElement('campaign-product-shot', CampaignProductShot);
+
+export default CampaignProductShot;
```
collin: Same here re: `export`

elements/campaign-product-shot/campaign-product-shot.js
daytonn: Yeah, you could just use invariant to enforce it: `invariant(data.product_shot_url, 'CampaignProductShot.handleRequestSuccess(data): data.product_shot_url is undefined')`

elements/campaign-brand-messaging/campaign-brand-messaging.js
daytonn: I would prefer to not log this kind of thing. It should be required and throw an error using invariant.

elements/campaign-brand-messaging/campaign-brand-messaging.test.js
```diff
@@ -0,0 +1,45 @@
...
+    element = document.createElement('campaign-brand-messaging');
+    element.setAttribute('src', src);
+
+    setImmediate(function () {
+      done();
+    });
+  });
+
+  it('renders an <campaign-brand-messaging>', function () {
+    assert.equal(element.tagName.toLowerCase(), 'campaign-brand-messaging');
```
kand: Not sure if this is the pattern elsewhere, but given we're using `expect(...)` syntax elsewhere in this test, we should use that here as well.

```
expect(element.tagName.toLowerCase()).to.equal('campaign-brand-messaging');
```

elements/campaign-brand-messaging/campaign-brand-messaging.js
collin: 👍

elements/campaign-product-shot/campaign-product-shot.js
collin: 👍

elements/campaign-product-shot/campaign-product-shot.js
```diff
@@ -0,0 +1,46 @@
...
+  BulbsHTMLElement,
+} from 'bulbs-elements/register';
+import { filterBadResponse, getResponseJSON } from 'bulbs-elements/util';
+import invariant from 'invariant';
+
+class CampaignProductShot extends BulbsHTMLElement {
+
+  handleRequestSuccess (data) {
+    if (!data.product_shot_url) {
+      console.log('Product shot url not set on this campaign');
```
kand: @collin are we doing `console.log` calls in this code for errors like this?

If so, include some campaign data with this message for easier debugging (e.g. the id or name).

collin: @daytonn's comment re using invariant applies here.

PR #155, opened 2016-09-16
extendDOM

collin:
@daytonn @kand @spra85 

We need to be able to set properties on the native dom element for our internal react components.

Use case for @jmelvnsn is:

``` js
let videoPlayer = document.createElement(`bulbs-video');
videoPlayer.play();
```

To get that on the dom node, we need this PR and to define:

``` js
class BulbsVideo extends BulbsElement {
  /* other metheds truncated */
  play () {
    /* code that plays video */
  }
}

Object.assign(BulbsVideo, {
  extendDOM: {
    play () {
      this.reactInstance.play();
    }
  }
});
```

This also renames `reactComponent` to `reactInstance` as that feels more explicit and correct.


spra85:
Seems straightforward.  @jmelvnsn are you needing this in order to call `play` on the `<video>` element based on what we talked about yesterday for the homepage player?  If so, I think we'd also need additional changes to other bulbs-elements, namely I think the `<video>` element needs to already be on the page underneath the image overlay for it to work as we expect, no?

collin:
I think that is correct, the jwplayer will have to be fully initialized and ready to call play on it's `<video>` element synchronously.

jmelvnsn:
@spra85 i believe so. what collin said tracks with my expectation

kand:
What is `extendDOM`? Is this coming from a React library?

kand:
👍

PR #158, opened 2016-09-20
bulbs-content

collin:
@kand


collin:
@daytonn @spra85 

`<bulbs-liveblog>` adds two two methods for cms-created content to end up on our sites after page load.

This provides an element to wrap our content in. Currently in only emits an event that can be used to initialize inline embeds (twitter, instagram, etc.)

Lo(ooooo)ng-term it would make sense to consolidate all those initializers into this element.

kand:
👍

collin:
@MelizzaP

collin:
@jmelvnsn @shawncook

spra85:
👍

PR #161, opened 2016-09-21
make meta cover hidden by default

benghaziboy:
@collin


collin:
This is related to what @jmelvnsn found?
Seems like a good plan 👍

benghaziboy:
yeah. I intended to get this out the first go through and was just a dunce.

PR #162, opened 2016-09-22
Bulbs Ellipsize

MelizzaP:
truncates text, implemented in the homepage and special coverage players

### DIFF
--- a/elements/bulbs-ellipsize/bulbs-ellipsize.js
+++ b/elements/bulbs-ellipsize/bulbs-ellipsize.js
@@ -10,8 +10,7 @@ class BulbsEllipsize extends BulbsHTMLElement {
 
     let lineCount = parseInt(this.getAttribute('line-count'));
 
-    debugger;
-    $(this).truncate({ lines: {lineCount} });
+    $(this).truncate({ lines: lineCount });
   }
 };
 
--- a/elements/bulbs-ellipsize/bulbs-ellipsize.js
+++ b/elements/bulbs-ellipsize/bulbs-ellipsize.js
@@ -1,6 +1,6 @@
 import { BulbsHTMLElement, registerElement } from 'bulbs-elements/register';
 import invariant from 'invariant';
-import Truncate from 'truncate.js';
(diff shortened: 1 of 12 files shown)


MelizzaP:
@collin @daytonn omg, this is finally ready for review

MelizzaP:
[Onion Special Coverage Backlog](https://trello.com/c/YMfvi3i6/667-onion-special-coverage-backlog)

elements/bulbs-ellipsize/bulbs-ellipsize.js
```diff
@@ -0,0 +1,22 @@
...
+      'BulbsEllipsize: Expects a line-count attribute');
+    if (!this.textContent) { return; }
+
+    this.truncate();
+  }
+
+  truncate () {
+    let lineCount = parseInt(this.getAttribute('line-count'), 10);
+
+    $(this).truncate({ lines: lineCount, ellipsis: '...' });
```
daytonn: I'm not sure we should put jQuery in these elements. Looks like you can do this vanilla js style: [https://www.npmjs.com/package/truncate.js#as-a-vanilla-javascript-object](https://www.npmjs.com/package/truncate.js#as-a-vanilla-javascript-object)

MelizzaP: any rational behind that? truncate.js is dependent on jquery already

daytonn: The rational is that jQuery is outdated in a code-base like bulbs-elements. There's simply no need to depend on it. I'm not sure what the internals look like for truncate.js but I would bet that it doesn't extend jQuery if it doesn't find it. In our other js code, we're not ready to let go of the almighty `$` but bulbs-elements is stuffed with new hotness. jQuery, in contrast, is old and busted.

MelizzaP: while I can always agree with new hotness, there is a bug in the vanilla js version of this. going with jquery.

PR #165, opened 2016-09-28
<script is="lazy-template"> (ready for review)

collin:
@mparent61 If you link in bulbs-elements through your `docker-compose.yml` we'll be able to start testing out `<script is="lazy-template">`

### DIFF
--- a/elements/lazy-template/lazy-template.js
+++ b/elements/lazy-template/lazy-template.js
@@ -30,10 +30,9 @@ class LazyTemplate extends BulbsHTMLScriptElement {
   }
 
   detachedCallback () {
-    else if (this.loadOn === 'in-view') {
+    if (this.loadOn === 'in-view') {
       this.tearDownLoadOnInView();
     }
-
   }
 
   setupLoadOnPageLoad () {
--- a/elements/lazy-template/lazy-template.js
+++ b/elements/lazy-template/lazy-template.js
@@ -1,0 +1,75 @@
+import { registerElement } from 'bulbs-elements/register';
+import { InViewMonitor } from 'bulbs-elements/util';
+import invariant from 'invariant';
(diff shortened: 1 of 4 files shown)


collin:
@kand @daytonn @spra85 @mparent61 

We're working on page speed and want to be able to ship html content without having the browser engine parse it.

According to the spec, this would be the time to use a `<template>` tag, but IE11 is real cool and parses the contents of `<template>` elements. So that's out for a while.

But browsers won't parse the contents of a `<script>` tag if it doesn't recognize the type.

enter:
```html
<script type="text/html">
```

Now, we want to customize this `<script>` tag to convert it's contents into real-live elements on the page at-will.

So we define a custom extension of the `<script>` tag, `is="lazy-template"`
```html
<script is="lazy-template" type="text/html">
```

There is another required attribute `load-on`


Page load loads when the window `load` event fires:
```html
 load-on="page-load"
```

In view loads the content when the <scirpt> template comes into the viewport.
```html
 load-on="in-view"
```

mparent61:
Awesome, can't wait to try this out.

elements/lazy-template/lazy-template.js
```diff
@@ -0,0 +1,75 @@
...
+      window.addEventListener('load', () => this.replaceWithContents());
+    }
+  }
+
+  setUpLoadOnInView () {
+    InViewMonitor.add(this);
+    this.addEventListener('enterviewport', this.handleEnterViewport);
+  }
+
+  tearDownLoadOnPageLoad () {
```
spra85: Do we want to leave this here if it's a no-op?

collin: Good catch.

daytonn:
👍

PR #167, opened 2016-10-01
Added optional attribute for embedded; when enabled, the advertising plugin is disabled

spra85:
@collin @daytonn @kand ... Queueing this up for a Monday review.

The purpose of the embedded attribute is to disable the JWPlayer advertising plugin for our own internal in-line article embedded players.

### DIFF
--- a/elements/bulbs-carousel/bulbs-carousel.test.js
+++ b/elements/bulbs-carousel/bulbs-carousel.test.js
@@ -98,7 +98,7 @@ describe('<bulbs-carousel>', () => {
     });
 
     context('called twice', () => {
-      it.only('does not create a second track', () => {
+      it('does not create a second track', () => {
         subject.createdCallback();
         expect(subject.querySelector('bulbs-carousel-track bulbs-carousel-track')).to.be.null;
       });
--- a/elements/bulbs-video/components/revealed.js
+++ b/elements/bulbs-video/components/revealed.js
@@ -97,9 +97,7 @@ export default class Revealed extends React.Component {
       videoMeta.player_options.defaultCaptions = true;
     }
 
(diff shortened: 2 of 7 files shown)


elements/bulbs-carousel/bulbs-carousel.test.js
```diff
@@ -98,7 +98,7 @@ describe('<bulbs-carousel>', () => {
     });
 
     context('called twice', () => {
-      it.only('does not create a second track', () => {
+      it('does not create a second track', () => {
```
spra85: This was causing the test suite to run only this test and skip all others.  So, yeah...

collin:
Looks good to me.

kand:
👍

elements/bulbs-video/components/revealed.js
```diff
@@ -97,6 +97,10 @@ export default class Revealed extends React.Component {
       videoMeta.player_options.defaultCaptions = true;
     }
 
+    if (this.props.embedded) {
+      videoMeta.player_options.embedded = true;
+    }
+
```
daytonn: I think `this.props.embedded` is already boolean so you could just `videoMeta.player_options.embeded = this.props.embeded`, right? Is there something else I'm missing?

daytonn:
other than that extra check, looks good 👍

elements/bulbs-video/components/revealed.js
daytonn: you could do the coercion trick if you're paranoid `videoMeta.player_options.embedded = !!this.props.embedded`

spra85: You're right, @daytonn.  Tests pass, so do the examples.  Updated it

PR #173, opened 2016-10-07
jw player embeded share tools

MelizzaP:
# Bug
The share tools embeded in the jwplayer were linking to window.location.href as opposed to the video url

# Fix
Added videohub reference id redirect to the jw player share tools

# Verification
In the upper right hand corner of the homepage video player click the share icon and then the facebook or twitter icon. In production it should link link to the homepage, in test it should link to the video url

## Onion
prod: http://www.theonion.com/
test: http://share-toolin.test.theonion.com

## Avclub
prod: http://www.avclub.com/
test: http://share-toolin.test.avclub.com


MelizzaP:
@collin @kand @daytonn

PR #177, opened 2016-10-11
(IN PROGRESS) Player integration for Freewheel video ad server

spra85:


elements/bulbs-video/components/revealed.js
```diff
@@ -156,13 +160,23 @@ export default class Revealed extends React.Component {
...
+  getProfValue () {
+    if (window.isMobile.any) {
+      return 'theonion_mobileweb_html5';
+    } else {
+      return 'theonion_desktop_html5';
+    }
+  }
 
+  vastUrl (videoMeta) {
+    let baseUrl = `http://${SETTINGS.FREEWHEEL_NETWORK_ID}v.fwmrm.net/ad/g/1?`;
```
spra85: @MelizzaP missing a `.` between the network id and the `v`.  Example (`http://1b656.v.fwmrm.net/ad/g/1?...`)

lib/bulbs-elements/settings.js
```diff
@@ -0,0 +1,31 @@
+let settings = {
+  base: {
+    BULBS_ELEMENTS_ONIONSTUDIOS_GA_ID: 'UA-223393-14',
+    BULBS_ELEMENTS_COMSCORE_ID: '6036328',
+    RESP: 'vmap1',
+  },
+  development: {
+    FREEWHEEL_NETWORK_ID: '111976',
```
spra85: I think these network IDs should actually be stored in each property's python settings and then passed into the the bulbs-element somehow, maybe even as a namespaced variable on `window`.  I don't believe the `NODE_ENV` gives you what you need to integrate into each of the bulbs properties, but I'd want to @collin to weigh in.

MelizzaP: my logic behind it was that they were the same for each site, so we wouldn't have to repeat them. And I just felt weird about putting them on the window.

elements/bulbs-video/components/revealed.js
```diff
@@ -8,11 +8,10 @@ import Comscore from '../plugins/comscore';
...
 import React, { PropTypes } from 'react';
 import invariant from 'invariant';
+import SETTINGS from 'bulbs-elements/settings';
 
-// FIXME: where should this be defined? Per-app?
-//  Or in some better sort of settings file here?
-global.BULBS_ELEMENTS_ONIONSTUDIOS_GA_ID = 'UA-223393-14';
-global.BULBS_ELEMENTS_COMSCORE_ID = '6036328';
+global.BULBS_ELEMENTS_ONIONSTUDIOS_GA_ID = SETTINGS.BULBS_ELEMENTS_ONIONSTUDIOS_GA_ID;
+global.BULBS_ELEMENTS_COMSCORE_ID = SETTINGS.BULBS_ELEMENTS_COMSCORE_ID;
```
collin: Rather than dump this on settings, I think we should just read it from `SETTINGS` every time we use these values.

lib/bulbs-elements/settings.js
collin: In the long-term I'd like to get `bulbs-elements` into `omni` and find some way to access the python settings at build time. Then we could have one settings file for everything.

But 1 settings.py location and 1 settings.js location beats having these settings strewn about the individual js files.

MelizzaP: if the end goal is to have all the settings in one `settings.py` file and we are pushing these settings to `window`. I'd argue, it might make sense to just put them in each sites' `settings.py` file. Having a `settings.js` and `settings.py` file just seems to add an unnecessary and possibly confusing extra layer of complication.

collin: 👍

elements/bulbs-video/components/revealed.js
```diff
@@ -156,13 +160,54 @@ export default class Revealed extends React.Component {
...
+    } else if (window.TARGETING.dfp_specialcoverage) {
+      let slug = window.TARGETING.dfp_specialcoverage;
+      return `specialcoverage_${slug}`;
+    } else {
+      return 'video';
+    }
+  }
+
+  getCsidValue (videoMeta) {
+    let deviceAcronym = this.setDeviceAcronym();
```
MelizzaP: I was trying to differentiate from values we are pulling in from other places the  vs values we are setting.

MelizzaP: but I suppose its really 6 of one half a dozen of the other

elements/bulbs-video/components/revealed.js
```diff
@@ -156,13 +160,54 @@ export default class Revealed extends React.Component {
...
+  setDeviceAcronym () {
+    if (window.isMobile.any) {
+      return 'm';
+    } else {
+      return 'd';
+    }
+  }
+
+  getSiteName () {
+    return window.location.host.split('.')[1];
```
MelizzaP: some videos have their `target_channel` set to `onion-studios`, do we want that value or should I just make it default to theonion?

MelizzaP:
[Add required query string parameters to the VAST ad request call](https://trello.com/c/0i2w8G0T/7-add-required-query-string-parameters-to-the-vast-ad-request-call)

elements/bulbs-video/components/revealed.js
```diff
@@ -156,30 +165,97 @@ export default class Revealed extends React.Component {
...
+      let slug = window.TARGETING.dfp_specialcoverage;
+      dfpSection = `specialcoverage_${slug}`;
     }
+    else {
+      dfpSection = 'video';
+    }
+    return dfpSection;
+  }
+
+  buildCsid (hostChannel) {
```
spra85: `buildCustomSiteSectionId`

elements/bulbs-video/components/revealed.js
```diff
@@ -156,30 +165,97 @@ export default class Revealed extends React.Component {
...
+  vastUrl (videoMeta) {
+    let hostChannel = videoMeta.hostChannel;
+    let videohubReferenceId = videoMeta.id;
+    let randomVideoPlayerNumber = videoMeta.vprn;
+    let vastTestId = this.vastTest(window.location.search);
+    let series = videoMeta.series_slug;
+    let campaignId = this.props.targetCampaignId;
+    let specialCoverage = this.props.targetSpecialCoverage;
+
+    let baseUrl = `http://${window.FREEWHEEL_NETWORK_ID}.v.fwmrm.net/ad/g/1?`;
```
spra85: This should be the freewheel network hash

elements/bulbs-video/components/revealed.js
```diff
@@ -156,30 +165,97 @@ export default class Revealed extends React.Component {
...
+
+    // required global params
```
spra85: We are missing the `nw` param, which should be `window.FREEWHEEL_NETWORK_ID`

elements/bulbs-video/components/revealed.js
```diff
@@ -156,30 +165,97 @@ export default class Revealed extends React.Component {
...
+    baseUrl += 'resp=' + 'vmap1';
+    baseUrl += '&prof=' + this.getProfValue();
+    baseUrl += '&csid=' + this.buildCsid(hostChannel);
+    baseUrl += '&caid=' + this.buildCaid(videohubReferenceId);
+    baseUrl += '&pvrn=' + this.cacheBuster();
+    baseUrl += '&vprn=' + randomVideoPlayerNumber;
+
+    // optional global param
+    if (vastTestId) { baseUrl += '&cana=' + vastTestId; }
```
spra85: Let's modify this line in `vastTest`: 

```
let vastId = this.parseParam('xgid', searchString);
```

So we make the parameter (which is ad placement id), let's call it `apid`

elements/bulbs-video/components/revealed.js
```diff
@@ -156,30 +165,97 @@ export default class Revealed extends React.Component {
...
+
+    // Key Values
+    baseUrl += ';&video_id=' + videohubReferenceId;
```
spra85: We shouldn't need the ampersand after the semi-colon, so `;&video_id` becomes `;video_id`

elements/bulbs-video/components/revealed.js
```diff
@@ -156,30 +165,97 @@ export default class Revealed extends React.Component {
...
+    baseUrl += '&channel_slug=' + videoMeta.channel_slug;
+    if (series) { baseUrl += '&series_slug=' + series; }
+    if (campaignId) { baseUrl += '&campaign_id=' + campaignId; }
+    if (specialCoverage) { baseUrl += '&special_coverage=' + specialCoverage; }
+
+    // Slot Params *Required Fields*
+    baseUrl += ';&slid=' + 'Preroll';
```
spra85: We shouldn't need the ampersand after the semi-colon, so `;&slid` becomes `;slid`

elements/bulbs-video/components/revealed.js
```diff
@@ -156,30 +171,96 @@ export default class Revealed extends React.Component {
...
+      dfpSection = `specialcoverage_${slug}`;
+    }
+    else {
+      dfpSection = 'video';
     }
+    return dfpSection;
+  }
+
+  buildCustomSiteSectionId (hostChannel) {
+    // format: <device acronym>.<site name>_<dfp section>_<host channel>
```
spra85: I think we can remove this comment, because that format is clear in the interpolation below

PR #178, opened 2016-10-11
Adding an optional attribute not to add a DFP pixel to the DOM

spra85:
@collin @MelizzaP @kand @daytonn ... Looking for a review on this one.  The purpose of this is to add an optional attribute, `no-pixel` to the campaign-display element that will not add an ad to the DOM unless desired.  The intent is to use this for the slide out navigation on the sites, particularly The Onion.

### DIFF
--- a/elements/bulbs-video/bulbs-video.js
+++ b/elements/bulbs-video/bulbs-video.js
@@ -1,7 +1,7 @@
 import React, { PropTypes } from 'react';
 import { registerReactElement } from 'bulbs-elements/register';
 import BulbsElement from 'bulbs-elements/bulbs-element';
-import { loadOnDemand } from 'bulbs-elements/util'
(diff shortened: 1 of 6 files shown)


collin:
I wonder if it would be wise to wrap `<campaign-display>` in `loadOnDemand` like we do now with `<bulbs-video>`?

spra85:
@collin, I think we'd ultimately probably want to do that as well.  But this covers a different scenario in that we never want to have the ad pixel regardless for the slide out nav

collin:
Gotcha :+1:

kand:
Can you add something about this feature to the [README](https://github.com/theonion/bulbs-elements/blob/22a684efedc6f0cecb9d8d2c85fd6130872450e1/elements/campaign-display/README.md)?

PR #179, opened 2016-10-18
<bulbs-page-state>

collin:


collin:
Some tests for `<bulbs-page>`

collin:
Running into some intermittent test failures on IE, 5 tests are failing in the full suite, but they pass when only their file is run.

Going to merge this in now so we can get the liveblog running, then will circle back to see whats up.

PR #180, opened 2016-10-19
Google Analytics Prefix in Lib

MelizzaP:


MelizzaP:
@collin @spra85

elements/bulbs-video/components/revealed.js
```diff
@@ -40,30 +40,35 @@ export default class Revealed extends React.Component {
...
+    let gaPrefix = makeGaPrefix();
+    let dimensions = {
+      'dimension1': targeting.target_channel || 'None',
+      'dimension2': targeting.target_series || 'None',
+      'dimension3': targeting.target_season || 'None',
+      'dimension4': targeting.target_video_id || 'None',
+      'dimension5': hostChannel,
+      'dimension6': specialCoverage,
+      'dimension7': true, // 'has_player' from old embed
+      'dimension8': this.props.autoplay || 'None',
```
collin: Could we retain the comments from the other side of the diff?

elements/bulbs-video/components/revealed.js
```diff
@@ -16,7 +17,6 @@ global.BULBS_ELEMENTS_COMSCORE_ID = '6036328';
 
 let prefixCount = 0;
 function makeGaPrefix () {
-  // ga demands tracker names be alphanumeric
   return `videoplayer${prefixCount++}`;
```
collin: Can the prefix counter stuff go into `prepGaEvent`?

So you pass a string into `prepGaEvent` and it makes it unique inside that function?

spra85: The GA prefix doesn't have any impact on tracking @MelizzaP.  It's just a way to namespace the tracker on the page.

lib/bulbs-elements/util/prep-ga-event.js
```diff
@@ -0,0 +1,18 @@
+export default function prepGaEvent (gaPrefix, gaId, dimensions = {}) {
```
spra85: I feel like the name of this function and file is misleading because you're prepping a GA tracker, not an event.  The GA tracker can do more then just sending events.  Reading this makes me think this should be `prepGaTracker` and `prep-ga-tracker.js`

elements/bulbs-video/components/revealed.js
```diff
@@ -247,7 +252,7 @@ export default class Revealed extends React.Component {
 
     player.setup(playerOptions);
 
-    GoogleAnalytics.init(player, videoMeta.gaPrefix);
+    GoogleAnalytics.init(player, videoMeta.sendAnalyticsEvent);
```
spra85: Not sure how I feel about passing a function by reference like this @collin / @MelizzaP.  Seems like the `sendAnalyticsEvent` logic using the GA extraction should just live in the GoogleAnalytics plugin.

MelizzaP: passing in the function keeps the code a little DRYer, we're not repetitively passing in the prefix and 'event'

lib/bulbs-elements/util/prep-ga-event.js
MelizzaP: since the function is returning a function that calls a ga event, I wanted to make that clear in the function name, maybe `prepGaEventTracker`?

spra85: Yeah, think that reads better to me

elements/bulbs-video/components/revealed.js
spra85: Going to defer to @collin.  I still don't think we should pass by reference like this but not enough to say that has to be changed.

lib/bulbs-elements/util/prep-ga-event.js
```diff
@@ -0,0 +1,18 @@
...
+  ga('create', gaId, 'auto', { name: gaPrefix });
+
+  for (let key in dimensions) {
+    ga(`${gaPrefix}.set`, key, dimensions[key]);
+  }
+
+  // return function that sends google analytics event
+  return (eventCategory, eventAction, eventLabel, eventValue) => {
```
collin: We could make this a little bit more generalized.

return a function that can do more than just `send`

```js
return (method, ...args) => {
  return global.ga(`${gaPrefix}.${method}`, ...args);
}
```

Then you get a function that can be used anywhere the global `ga` function would be used, but it will get prefixed.

elements/bulbs-video/components/revealed.js
collin: I'm pro passing in the tracking function into the plugin. But per my other comment I think both of these should be valid:

```js
GoogleAnalytics.init(player, global.ga);
```

or

```js
GoogleAnalytics.init(player, prefixedGa);
```

MelizzaP:
@collin

elements/bulbs-video/plugins/google-analytics.test.js
```diff
@@ -68,22 +69,22 @@ describe('Google Analytics', () => {
       });
 
       it('sends a "play" event', () => {
-        expect(global.ga).to.have.been.calledWith(
-          'videoplayer0.send', 'event', 'Video:The Onion', 'play', 'http://www.theonion.com/r/4053'
+        expect(gaTrackerAction).to.have.been.calledWith(
+          'event', 'Video:The Onion', 'play', 'http://www.theonion.com/r/4053'
```
MelizzaP: doh

elements/bulbs-video/components/revealed.js
```diff
@@ -202,7 +202,7 @@ export default class Revealed extends React.Component {
   }
 
   makeVideoPlayer (element, videoMeta) {
-    element.id = videoMeta.gaPrefix;
+    element.id = jwPlayerIdCounter++;
```
spra85: Possibly a mute point, but we really should be prefixing this instead of generically using integer ids, which I feel like could easily collide with other elements eventually.  Instead of ids `0`, `1`, `2`... it makes more sense to do something like `jw-player-0`, `jw-player-1`, `jw-player-2`

collin: Good call.

spra85: @MelizzaP can you make this quick change as a new PR?

MelizzaP: on it

PR #182, opened 2016-10-25
append prefix to jw-player element id

MelizzaP:
test link: **http://jw-player-test.test.theonion.com/interactive/american-voter**

### DIFF
--- a/elements/bulbs-video/components/revealed.js
+++ b/elements/bulbs-video/components/revealed.js
@@ -202,7 +202,7 @@ export default class Revealed extends React.Component {
   }
 
   makeVideoPlayer (element, videoMeta) {
-    element.id = jwPlayerIdCounter++;
+    element.id = `jw-player-${jwPlayerIdCounter++}`;
     let player = global.jwplayer(element);
 
     player.videoMeta = videoMeta;


MelizzaP:
@spra85

elements/bulbs-video/components/revealed.js
```diff
@@ -202,7 +202,7 @@ export default class Revealed extends React.Component {
   }
 
   makeVideoPlayer (element, videoMeta) {
-    element.id = jwPlayerIdCounter++;
+    element.id = `jw-player-${jwPlayerIdCounter++}`;
```
spra85: @MelizzaP, can you confirm this also increments the original value?  If there are 3 players on the same page, I wanted to confirm the assignment works.  Maybe we can point a test link at Collin's american voters page as a perfect recent example. 

e.g. 
```
let jwPlayerIdCounter=0;
foo = `jw-player-${jwPlayerIdCounter++}`;
// value foo is `jw-player-1`, but is jwPlayerIdCounter the variable still 0?
```

spra85: It does require it, but IDs are supposed to be unique on the page and I don't know the internals of how the jwplayer library handles it.  I think their library maintainers are likely safely assuming the ID handed to them is unique on the page.

MelizzaP: tested it locally with the onion homepage player and it does, not sure if that will change with multiple players, I'll ping you when the link is up

MelizzaP:
@kand @collin can I get one thumb on this?

collin:
👍

PR #186, opened 2016-10-25
a utility that tracks whether or not a human has interected yet

collin:
```js
import Human from 'bulbs-elements/util';

Human.hasInteracted;
=> either true | false
```

`Human.hasInteracted` is set on cusor, keyboard, mousewheel, touch, and similar events.

@kand @spra85 @MelizzaP 

(micro pr from american voter)


lib/bulbs-elements/util/index.js
```diff
@@ -1,27 +1,29 @@
+import Human from './human';
```
collin: I sorted these alphabetically.

PR #190, opened 2016-10-26
Unmount bulbs-video

jmelvnsn:
This fixes an issue where bulbs-video elements were still playing after they had been unmounted. An example can be viewed here: https://github.com/theonion/omni/pull/286

@collin @kand @MelizzaP

### DIFF
--- a/elements/bulbs-video/components/revealed.js
+++ b/elements/bulbs-video/components/revealed.js
@@ -18,6 +18,13 @@ global.BULBS_ELEMENTS_COMSCORE_ID = '6036328';
 let jwPlayerIdCounter = 0;
 
 export default class Revealed extends React.Component {
+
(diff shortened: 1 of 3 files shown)


elements/bulbs-video/components/revealed.js
```diff
@@ -18,6 +18,13 @@ global.BULBS_ELEMENTS_COMSCORE_ID = '6036328';
 let jwPlayerIdCounter = 0;
 
 export default class Revealed extends React.Component {
+
+  componentWillUnmount () {
```
collin: Can we get a test that `componentWillMount` calls `this.player.stop`?

And don't need the extra blank lines.

jmelvnsn: Yeah. Good call

PR #192, opened 2016-10-31
AD block Rail Player

jmelvnsn:
This addition to bulbs-elements adds an ad-check util and is available to other components. This adds a query param to our rail video request if ad block returns true and we update the content 
here: 

https://github.com/theonion/omni/pull/345

@spra85 @collin @mparent61


jmelvnsn:
This addition to bulbs-elements adds an ad-check util and is available to other components as well. We also a query param to our rail video request if ad block returns true and update that content here: https://github.com/theonion/omni/pull/345

spra85:
Approved pending tests @jmelvnsn

PR #193, opened 2016-11-02
Fix double audio bug

collin:
fixes a bug where there were two audio tracks playing at once

Also fixes an issue with the video carousel where it skipped ot the next clip when an ad finished, rather than when the video finished

# To Reproduce

Go to: http://www.avclub.com

Play an item in the video carousel. Wait for the ad to finish. The carousel will advance to the next video without playing the one you requested, and there will be multiple audio streams playing on top of each other.

# To Test

Go to: http://fix-double-video.test.avclub.com

Play an item in the video carousel. Wait for the ad to finish. The content should start and only one audio track should play.

Scrub to almost the end of the video and wait for the video to complete. The next item in the carousel should start playing.

### DIFF
--- a/elements/bulbs-video/components/revealed.js
+++ b/elements/bulbs-video/components/revealed.js
@@ -19,9 +19,9 @@ global.BULBS_ELEMENTS_COMSCORE_ID = '6036328';
 let jwPlayerIdCounter = 0;
 
 export default class Revealed extends React.Component {
-
-  componentWillUnmount () {
-    this.player.stop();
+  constructor (props) {
+    super(props);
+    this.forwardJWEvent = this.forwardJWEvent.bind(this);
   }
 
   componentDidMount () {
@@ -106,6 +106,10 @@ export default class Revealed extends React.Component {
     videoMeta.player_options.embedded = this.props.embedded;
 
     this.makeVideoPlayer(this.refs.videoContainer, videoMeta);
+  }
+
+  componentWillUnmount () {
+    this.player.remove();
   }
 
   extractSources (sources) {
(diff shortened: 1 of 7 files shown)


elements/bulbs-video/components/revealed.js
```diff
@@ -108,6 +108,10 @@ export default class Revealed extends React.Component {
     this.makeVideoPlayer(this.refs.videoContainer, videoMeta);
   }
 
+  componentWillUnmount () {
+    this.player.remove();
```
collin: Stopping is not enough. The player must be completely removed.

elements/bulbs-video/components/revealed.js
```diff
@@ -256,11 +260,16 @@ export default class Revealed extends React.Component {
...
+  }
+
+  forwardJWEvent (event) {
+    this.refs.videoViewport.dispatchEvent(new CustomEvent(`jw-${event.type}`));
   }
 
   render () {
     return (
-      <div className='bulbs-video-viewport'>
+      <div className='bulbs-video-viewport' ref='videoViewport'>
```
collin: jwplayer seems to discard/replace the element you tell it to use, so I needed a hook to this wrapping element to dispatch events from.

elements/bulbs-video/elements/video-carousel.js
```diff
@@ -57,7 +57,7 @@ class BulbsVideoCarousel extends BulbsHTMLElement {
       '<bulbs-video-carousel> MUST contain a <bulbs-carousel>'
     );
 
-    this.videoPlayer.addEventListener('ended', this.playerEnded = this.playerEnded.bind(this), true);
+    this.videoPlayer.addEventListener('jw-complete', this.playerEnded = this.playerEnded.bind(this), true);
```
collin: The video `ended` fires for ads and content.

BulbsVideo now forwards the jwplayer `complete` event as `jw-complete`, this only fires when content completes, not an ad.

elements/bulbs-video/components/revealed.js
```diff
@@ -256,11 +260,16 @@ export default class Revealed extends React.Component {
     GoogleAnalytics.init(this.player, videoMeta.gaTrackerAction);
     Comscore.init(this.player, global.BULBS_ELEMENTS_COMSCORE_ID, videoMeta.player_options.comscore.metadata);
 
+    this.player.on('complete', this.forwardJWEvent);
```
spra85: @collin, which part of the PR dictates that it only forwards the event when content ends and not ads?

collin:
For completeness: jwplayer emits separate events for ads.
```js
jwplayer().on('adComplete')
```

PR #194, opened 2016-11-04
In view page tracking

collin:
This change emits

`progresschange`
`pagestart`
`pageend`

events on elements that are tracked by the `InViewMonitor` utility.

This is used by the `<bulbs-page>` element to determine when to fire analytics events and alter the address bar with the pushstate api.

This duplicates part of the capabilities of the `<bulbs-reading-list>` and it would be wise to migrate the reading list to use `InViewMonitor` for these events. `InViewMonitor` batches calls to `getBoundingClientRect` which prevents inefficient 'layout thrashing'

### DIFF
--- a/elements/bulbs-dfp/bulbs-dfp.js
+++ b/elements/bulbs-dfp/bulbs-dfp.js
@@ -149,7 +149,7 @@ export default class BulbsDfp extends BulbsHTMLElement {
   }
 
   get isViewable () {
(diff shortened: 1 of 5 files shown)


collin:
@spra85 @kand @MelizzaP

collin:
We're experiencing Browserstack timeouts in travis. I ran these tests manually on browserstack.com

PR #195, opened 2016-11-04
Tests for scroll locking

collin:
This adds a utility that locks the scroll position to a given element.

`util.ScrollLock.lockToElement`

It stops locking scroll position when the user interacts with the browser, by scrolling, typing, touching, etc.

It adjust the scroll position on every animation frame, this mitigates the content-reflow that would otherwise bump the scroll position.

@kand @MelizzaP @spra85

### DIFF
--- a/elements/bulbs-dfp/bulbs-dfp.test.js
+++ b/elements/bulbs-dfp/bulbs-dfp.test.js
@@ -241,9 +241,10 @@ describe('<div is="bulbs-dfp">', () => {
     beforeEach((done) => {
       element.style.height = '100px';
       element.style.width = '100px';
-      element.style.position = 'absolute';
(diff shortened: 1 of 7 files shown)


spra85:
Seems straightforward @collin, but can you describe the scenario(s) you're needing to reach for this utility?

collin:
Yeah, we're using this with

```html
<bulbs-page
  pushstate-url="/page-url"
  lock-scroll-on-load
>
```

Was already used in http://www.theonion.com/interactive/american-voter.

With those attributes set, we scroll to the element with a `pushstate-url` matching the current window.location and stick there until the user scrolls. Locking in this fashion mitigates any scroll issue as assets above the intended element load in and reflow the layout.

PR #198, opened 2016-11-07
Lazy load ads

spra85:
This change is coupled with the updates to the ads manager here (https://github.com/theonion/bulbs-public-ads-manager/pull/20) and makes the call to refresh the ad slot using the new method exposed by the ads manager PR.

Test link below (http://lazy-load-ads.test.clickhole.com), items for regression test: 
- Homepage ads
- Campaign pixels
- Article ads
- `inread` ads in article body


collin:
We've started seeing some test issues with IE11 that seem to only fail when running the whole suite.

We have been been manually testing for safari/Firefox/chrome/ie to see if they pass. I plan to look in on this today while babysitting the liveblog.

spra85:
@collin noticed this too.  I'm assuming this won't hold up the deploy of this?

PR #203, opened 2016-11-09
Removing a lot of the event chatter to measure refresh/viewport check

spra85:
This PR removes a lot of the event tracking noise that pushed us over our GA monthly hit allotment last month.  Preserves an event only for when an ad refresh is triggered.

Test link: 
http://less-event-noise.test.theonion.com/


collin:
👍 

new career line item:

* 2016: Broke GA

PR #206, opened 2016-11-15
Bulbs responsive nav

jmelvnsn:
@collin @spra85 @kand @shawncook @MelizzaP 

TESTLINK: http://bulbs-header-test.test.theonion.com/?dfptest=pagepush


elements/bulbs-header/bulbs-header-test.js
```diff
@@ -0,0 +1,47 @@
...
+
+describe('<bulbs-header>', () => {
+  let header;
+  let masthead;
+  let responsiveNav;
+
+  beforeEach((done) => {
+    header = document.createElement('bulbs-header');
+    header.innerHTML = `
+      <bulbs-header-mastead></bulbs-header-mastead>
```
spra85: Is the `masthead` variable just not used in these tests, because `<bulbs-header-masthead>` is spelled wrong here.

elements/bulbs-header/bulbs-header.js
```diff
@@ -0,0 +1,65 @@
...
+import './bulbs-header-masthead';
+import './bulbs-header-responsive-nav';
+import './bulbs-header.scss';
+
+import { InViewMonitor } from 'bulbs-elements/util';
+
+class BulbsHeader extends BulbsHTMLElement {
+  createdCallback () {
+    this.mastheads = this.getElementsByTagName('bulbs-header-masthead');
+    this.responsiveNavs = this.getElementsByTagName('bulbs-header-responsive-nav');
```
spra85: Is there ever going to be more than one of either of these?  Wondering if we should do the `[0]` index look up here and eliminate those prop accessors below

collin: The use of `getElementsByTagName` and the `[0]` lookup is defensive. `getElementsByTagName` returns a live collection that updates as the dom updates.

Always looking at the head of that collection means you can break the reference to the actual internal elements without breaking the element as a whole.

But it could be too defensive, this would also work:

```js
this.masthead = this.querySelector('bulbs-header-masthead');
this.responsiveNav = this.querySelector('bulbs-header-responsive-nav');
```

And then delete the getters.

spra85: @collin, defer to you on if we should make the change, just seemed unnecessary to have the getters when I reviewed.

collin: I'd say we kill the use of `getElementsByTagName` and the getters if it's unclear what they're doing.

elements/bulbs-header/bulbs-header-responsive-nav.js
```diff
@@ -0,0 +1,9 @@
+import { registerElement, BulbsHTMLElement } from 'bulbs-elements/register';
+
+class BulbsHeaderResponsiveNav extends BulbsHTMLElement {
+
```
jmelvnsn: Chatted with Collin. We are going to keep it.

elements/bulbs-header/bulbs-header.js
jmelvnsn: ++

PR #208, opened 2016-11-17
allow for a series url prefix in the video meta

collin:
This is a compromise to get us around a sticky situation. Our properties don't have full video data at time of page render.  This leads us to do quite a bit of json fetching from onionstudios to complete that data on the front end.

The video json data has a series url that points to onionstudios.com.

Now our properties are starting to host their own series pages. But that url isn't available in the video meta element.

We're letting the property define the url prefix that is to be mashed together with the series slug.

@kand @spra85 (@mparent61 getting you in here because we discussed this previously)

This will be used as:

```html
<bulbs-video-meta relative-series-link-prefix="/series">
```

### DIFF
--- a/elements/bulbs-video/elements/meta/components/root.js
+++ b/elements/bulbs-video/elements/meta/components/root.js
@@ -8,16 +8,30 @@ export default function VideoMetaRoot (props) {
   }
 
   let seriesElement;
-  let seriesNameElement = <h2 className='bulbs-video-meta-series-name'>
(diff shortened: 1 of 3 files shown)


collin:
[series type in the right rail above the title should now link to onion video series page  ex. http://www.theonion.com/series/onion-film-standard](https://trello.com/c/BGiFvFhD/995-series-type-in-the-right-rail-above-the-title-should-now-link-to-onion-video-series-page-ex-http-www-theonion-com-series-onion-f)

PR #210, opened 2016-11-28
style first carousel item properly on first play

kand:
This styles the first item in the carousel properly on first video play.

**Release Type**: _patch_
No major features, only a fix.


kand:
[Homepage Player First Video Play should also show grey'd out state with border](https://trello.com/c/eL5YtapI/960-homepage-player-first-video-play-should-also-show-grey-d-out-state-with-border)

PR #211, opened 2016-12-07
reading list with responsive header interactions

kand:
Refactor pinned element and remove reading list specific pinning logic to allow site-specific usage to pinned element in reading lists.

### Release Type: _major_
`bulbs-pinned-element` is not backwards compatible.

### Breaking

1. `bulbs-reading-list-menu` no longer has sticky logic out of the box. Sites should determine themselves if the menu should be pinned, in which case the `bulbs-pinned-element` should be used to wrap it. Related attributes have been removed and should be removed on all usages.

### New

1. `getRoundedBoundingClientRect` can be used to get a `Math.ceil` rounded version of an elements `getBoundingClientRect`. Will avoid some measurement comparison issues in some browsers.

### Updated

1. `bulbs-pinned-element` now only takes `offset-top-px`.

### Known Issues

1. `bulbs-pinned-element` example does not currently work. Will have to modify the logic in the future to allow for a scroll container other than `window`.

### DIFF
--- a/elements/bulbs-pinned-element/bulbs-pinned-element-examples.js
+++ b/elements/bulbs-pinned-element/bulbs-pinned-element-examples.js
@@ -4,10 +4,17 @@ let examples = {
     'Basic Example': {
       render: () => {
         return `
-          <bulbs-pinned-element
-            attribute="value"
-          >
-          </bulbs-pinned-element>
+          <div
+              style="
+                background-color: blue;
+                height: 1000px;
+                width: 200px;
+              "
+              >
+            <bulbs-pinned-element scroll-container-selector=".examples-index-pane">
+              <div>this is pinned content!</div>
+            </bulbs-pinned-element>
+          </div>
         `;
       },
     },
--- a/elements/bulbs-pinned-element/bulbs-pinned-element.js
+++ b/elements/bulbs-pinned-element/bulbs-pinned-element.js
@@ -6,7 +6,8 @@ import {
 } from 'bulbs-elements/register';
 import {
   moveChildren,
-  getScrollOffset
+  getScrollOffset,
+  InViewMonitor
 } from 'bulbs-elements/util';
 import './bulbs-pinned-element.scss';
 
@@ -48,19 +49,22 @@ export default class BulbsPinnedElement extends BulbsHTMLElement {
   }
 
   positionCar () {
+    const boundingRects = this.getBoundingRects();
+
     if(!this.animationRequest) {
       this.animationRequest = requestAnimationFrame(() => {
         this.animationRequest = null;
 
-        const boundingRects = this.getBoundingRects();
-
-        this.style.height = `${boundingRects.parent.height}px`;
+        this.style.height = `${boundingRects.parent.height - Math.abs(boundingRects.parent.top - boundingRects.rail.top)}px`;
         this.style.width = `${boundingRects.parent.width}px`;
 
-        if(this.isScrollingDown()) {
-          this.handleScrollDown(boundingRects);
-        } else {
-          this.handleScrollUp(boundingRects);
+        if (InViewMonitor.isElementInViewport(this, boundingRects.rail)) {
+
+          if(this.isScrollingDown()) {
+            this.handleScrollDown(boundingRects);
+          } else {
+            this.handleScrollUp(boundingRects);
+          }
         }
       });
     }
@@ -99,10 +103,9 @@ export default class BulbsPinnedElement extends BulbsHTMLElement {
 
   pinToRailTop () {
     this.car.classList.remove('pinned', 'pinned-bottom');
-    this.car.classList.add('pinned-top');
 
     this.car.style.bottom = 'initial';
-    this.car.style.top = 'initial';
+    this.car.style.top = 0;
   }
 
   pinToRailBottom () {
--- a/elements/bulbs-pinned-element/bulbs-pinned-element.js
+++ b/elements/bulbs-pinned-element/bulbs-pinned-element.js
@@ -127,7 +127,7 @@ export default class BulbsPinnedElement extends BulbsHTMLElement {
   resetCarPosition () {
     this.car.classList.remove('pinned', 'pinned-bottom');
 
-    this.car.style.bottom = 'initial';
+    this.car.style.bottom = '';
     this.car.style.top = 0;
   }
 
@@ -136,14 +136,14 @@ export default class BulbsPinnedElement extends BulbsHTMLElement {
     this.car.classList.add('pinned-bottom');
 
     this.car.style.bottom = 0;
-    this.car.style.top = 'initial';
+    this.car.style.top = '';
   }
 
   pinToWindow () {
     this.car.classList.remove('pinned-bottom');
     this.car.classList.add('pinned');
 
-    this.car.style.bottom = 'initial';
+    this.car.style.bottom = '';
     this.car.style.top = `${this.topOffsetAdjustment}px`;
   }
 }
--- a/elements/bulbs-pinned-element/bulbs-pinned-element.js
+++ b/elements/bulbs-pinned-element/bulbs-pinned-element.js
@@ -6,15 +6,10 @@ import './bulbs-pinned-element.scss';
 export default class BulbsPinnedElement extends BulbsHTMLElement {
 
   attachedCallback () {
-    invariant(this.hasAttribute('pinned-to'), '<bulbs-pinned-element pinned-to=".selector">: a pinned-to selector is required');
 
-    const selector = this.getAttribute('pinned-to');
-    const element = document.querySelector(selector);
-    this.topOffsetAdjustment = this.getAttribute('offset-top') || 0;
+    this.topOffsetAdjustment = parseInt(this.getAttribute('offset-top-px') || 0, 10);
     this.lastPosition = 0;
     this.animationRequest = null;
-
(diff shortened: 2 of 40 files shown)


kand:
@shawncook @MelizzaP getting responsive header working so I can test locally. Didn't want to break any changes in your branch so made this one, let me know if we can merge this in.

kand:
[basic reading list implementation](https://trello.com/c/LYH9p5LW/3-basic-reading-list-implementation)

elements/bulbs-pinned-element/bulbs-pinned-element.js
```diff
@@ -1,34 +1,46 @@
-import { registerElement, BulbsHTMLElement } from 'bulbs-elements/register';
 import invariant from 'invariant';
-import { getScrollOffset, InViewMonitor } from 'bulbs-elements/util';
+
```
kand: @collin @MelizzaP @shawncook switched up the `bulbs-pinned-element` to move the content to the inside of a `car` on a `rail` and move only in respect to that `rail`. My assumption is that the rail will always match up with the content height that the rail should be following. This gives us a bit more control over how the pinned element's container interacts with the content inside of it.

Working on tests now, but in the meantime please review my `bulbs-pinned-element` changes in isolation from the rest of the PR. Let me know, at least conceptually, if you agree with these changes or want to use a different strategy.

kand:
[remove reading list menu pinning logic in bulbs-elements](https://trello.com/c/hXFUsHJx/8-remove-reading-list-menu-pinning-logic-in-bulbs-elements)

elements/bulbs-pinned-element/bulbs-pinned-element.js
```diff
@@ -1,128 +1,133 @@
...
-    invariant(this.hasAttribute('pinned-to'), '<bulbs-pinned-element pinned-to=".selector">: a pinned-to selector is required');
 
-    const selector = this.getAttribute('pinned-to');
-    const element = document.querySelector(selector);
+    this.topOffsetAdjustment = parseInt(this.getAttribute('offset-top-px') || 0, 10);
     this.lastPosition = 0;
     this.animationRequest = null;
 
-    invariant(element, `<bulbs-pinned-element pinned-to=".selector">: no element with the selector "${selector}" is in the DOM`);
+    let car = this.querySelector('bulbs-pinned-element-car');
```
MelizzaP: can we get throw an error when this element is missing? or update the README? preferably both

kand: This child element is provided by the `bulbs-pinned-element` itself in the following lines. Here I'm mirroring the defensive logic contained in some other components to keep a reference to existing components if the element is unmounted and remounted by external logic.

I will update the readme for this component before review and release.

elements/bulbs-pinned-element/bulbs-pinned-element-examples.js
```diff
@@ -4,10 +4,17 @@ let examples = {
     'Basic Example': {
       render: () => {
         return `
-          <bulbs-pinned-element
-            attribute="value"
-          >
-          </bulbs-pinned-element>
+          <div
```
kand: This example doesn't work at this point since pinned element tracks on window scroll instead of in relation to a relative scrolling container. This will be logged as a known issue.

elements/bulbs-pinned-element/bulbs-pinned-element.js
```diff
@@ -1,128 +1,150 @@
...
     if(!this.animationRequest) {
       this.animationRequest = requestAnimationFrame(() => {
         this.animationRequest = null;
 
-        if(!InViewMonitor.isElementInViewport(this, this.getBoundingClientRect())) { return; }
+        const boundingRects = this.getBoundingRects();
 
-        let elementPinnedTo = this.getElementPinnedTo();
-        const boundingRects = this.getBoundingRects(elementPinnedTo);
+        this.style.height = `${boundingRects.parent.height - Math.abs(boundingRects.parent.top - boundingRects.rail.top)}px`;
```
kand: This calculation allows the rail to share space in its parent with other elements and still be able to stick at the right time.

elements/bulbs-pinned-element/bulbs-pinned-element.js
```diff
@@ -1,128 +1,150 @@
...
+        this.style.width = `${boundingRects.parent.width}px`;
 
-        if(this.isScrollingDown()) {
-          this.handleScrollDown(elementPinnedTo, boundingRects);
+        if (this.hasNewRailHeight(boundingRects)) {
```
kand: Allows adjustments when the rail changes heights, i.e. new content has loaded into a reading list and now the sidebar is taller.

elements/bulbs-pinned-element/bulbs-pinned-element.js
```diff
@@ -1,128 +1,150 @@
...
+          this.resetCarPosition();
         }
-        else {
-          this.handleScrollUp(boundingRects);
+
+        if (this.isInView(boundingRects)) {
```
kand: Prevents handlers from firing when rail isn't in view.

elements/bulbs-pinned-element/bulbs-pinned-element.scss
```diff
@@ -1,3 +1,19 @@
+
```
kand: Added some very basic styling to get the pinned element working.

elements/bulbs-reading-list/components/bulbs-reading-list-menu.js
```diff
@@ -5,11 +5,6 @@ import {
 } from 'bulbs-elements/register';
 
 class BulbsReadingListMenu extends BulbsHTMLElement {
-  attachedCallback () {
-    invariant(this.getAttribute('reading-list-id'), '<bulbs-reading-list-menu> requires a "reading-list-id" attribute');
-    invariant(this.getAttribute('fixed-menu-min-width'), '<bulbs-reading-list-menu> requires a "fixed-menu-min-width" attribute');
-    invariant(this.getAttribute('pinned-container-selector'), '<bulbs-reading-list-menu> requires a "pinned-container-selector" attribute');
-  }
```
kand: None of this is needed now that the menu doesn't have an out of the box concept of sticking.

elements/bulbs-reading-list/lib/reading-list-menu.js
```diff
@@ -8,32 +8,13 @@ export default class ReadingListMenu {
     invariant(element, 'new ReadingListMenu(element, dispatcher): element is undefined');
     invariant(dispatcher, 'new ReadingListMenu(element, dispatcher): dispatcher is undefined');
 
-    this.element = element;
```
kand: All the reading list menu specific sticky logic has been removed.

lib/bulbs-elements/util/dimensions.js
```diff
@@ -38,6 +38,26 @@ export function getScrollOffset () {
   return { x, y };
 }
 
+export function getRoundedBoundingClientRect (element) {
```
kand: Need this utility to get around firefox having too much precision with these pixel values. Prevents problems trying to figure out when bounding box values are equal.

lib/bulbs-elements/util/dimensions.js
```diff
@@ -38,6 +38,26 @@ export function getScrollOffset () {
...
+  // firefox has too much precision and will create unexpected behavior when
+  //  comparing bounding rect pixel values. ensure the math always works out the
+  //  same by rounding. who's rendering 10000ths of a pixel anyways?
+  let rect = element.getBoundingClientRect();
+
+  // NOTE: cannot copy ClientRect in a normal way because it's a special object
+  return [
```
kand: A wonky way of doing this, but `ClientRect` cannot be copied in the normal ways because it's a special object. Let me know if you have a cleaner way of doing this.

package.json
```diff
@@ -135,5 +135,9 @@
     "yo": "^1.6.0",
     "scroll-into-view": "^1.7.1",
     "waypoints": "^4.0.1"
+  },
+  "devDependencies": {
+    "chai-spies": "^0.7.1",
+    "mock-raf": "^0.1.0"
```
kand: Gives control of `requestAnimationFrame` to tests.

elements/bulbs-reading-list/lib/reading-list-articles.test.js
```diff
@@ -38,7 +38,7 @@ describe('ReadingListArticles', () => {
         href,
         partialUrl,
         title,
-        contentAnalyticsDimensions,
+        contentAnalyticsDimensions
```
kand: 👍 Fixed this and other linting problems.

kand:
@spra85 will release as a `major` version.

lib/bulbs-elements/util/dimensions.js
collin: That sounds right.

elements/bulbs-reading-list/components/bulbs-reading-list-menu.js
collin: I think we can kill the entire file if this element is doing nothing.

kand:
@collin @MelizzaP @spra85 going to release this without fixing remaining travis tests. These tests don't have anything to do with my changes as far as I can tell.

MelizzaP:
😭, they have been failing for a bit. There is a ticket to fix them, still frustrating though

PR #214, opened 2016-12-13
fix dissapearing menu behavior in chrome

MelizzaP:
test link: http://clickhole-rl-qa.test.clickhole.com/article/6-pieces-meat-i-have-seen-held-and-one-case-gone-v-4972


elements/bulbs-reading-list/lib/reading-list-menu.js
```diff
@@ -49,12 +49,6 @@ export default class ReadingListMenu {
     return this.menuItems[0];
   }
 
-  setMenuPosition () {
```
MelizzaP: naw, must've got lost in refactoring

PR #215, opened 2016-12-13
Update revealed.js to include shreUrl as an attribute of bulbs-video

jmelvnsn:
@collin @kand @spra85

### DIFF
--- a/elements/bulbs-video/components/revealed.js
+++ b/elements/bulbs-video/components/revealed.js
@@ -72,7 +72,7 @@ export default class Revealed extends React.Component {
     let videoMeta = Object.assign({}, this.props.video);
     videoMeta.hostChannel = hostChannel;
     videoMeta.gaTrackerAction = gaTrackerAction;
-    videoMeta.player_options.shareUrl = this.props.shareUrl;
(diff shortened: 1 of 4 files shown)


elements/bulbs-video/components/revealed.js
```diff
@@ -72,7 +72,7 @@ export default class Revealed extends React.Component {
     let videoMeta = Object.assign({}, this.props.video);
     videoMeta.hostChannel = hostChannel;
     videoMeta.gaTrackerAction = gaTrackerAction;
-    videoMeta.player_options.shareUrl = `${window.location.href}/v/${videoMeta.id}`;
+    videoMeta.player_options.shareUrl = this.props.shareUrl;
```
spra85: Can we still preserve the default functionality for this PR, even if it's just a temporary thing in case we miss any refs that now need to pass in the `shareUrl` attribute?  So, basically advocating for: 

```
this.props.shareUrl || `${window.location.href}/v/${videoMeta.id}`;
```

collin: :+1: to that.

jmelvnsn: Cool

PR #217, opened 2016-12-15
REAL autoplay/pause in view

collin:
@MelizzaP this is the simplest way of doing this.

@spra85 there was discussion of having some sort of smarter video control service that would live on the property. Something like this ?

```js
// if we did something like this there would NOT be handlers for the 'enterviewport' and 'exitviewport' within Revealed
function playOneAndOnlyOneVideo () { /* magic happens */ }
$(document.body).on('enterviewport', 'bulbs-video', playOneAndOnlyOneVideo);
$(document.body).on('exitviewport', 'bulbs-video', playOneAndOnlyOneVideo);
``` 

Using event delegation there: https://learn.jquery.com/events/event-delegation/

## Update
used this branch to make an actual play in view/pause out of view functionality

## Test link
http://ch-rl-pause-vids.test.clickhole.com/video/share-video-your-sisters-wall-let-her-know-youre-n-5249

Videos should pause out of view, and should not play after you manually pause them


MelizzaP:
@collin @spra85 @kand this is ready for review, still have some safari tests failing in polls. I believe we're going to fix those in 2017

elements/bulbs-video/components/revealed.js
```diff
@@ -258,6 +260,17 @@ export default class Revealed extends React.Component {
       };
     }
 
+    if (typeof this.props.autoplayInView === 'string') {
+      let videoViewport = this.refs.videoViewport;
```
spra85: Can we move this block into a separate function?

MelizzaP:
[Video should play when in view, not on load](https://trello.com/c/rHZIANkB/125-video-should-play-when-in-view-not-on-load)

collin:
Yay, so much learning.

elements/bulbs-video/components/revealed.js
```diff
@@ -274,6 +303,16 @@ export default class Revealed extends React.Component {
     }
   }
 
+  handlePauseEvent (reason) {
+    if (reason.pauseReason === 'external') {
+      return;
+    }
+    else if (reason.pauseReason === 'interaction') {
```
spra85: Sweet.

MelizzaP: yeah, unfortunately this functionality won't work with ads. `jwplayer` does not attach a pause reason to those

spra85: Can we make the assumption that the absence of a `pauseReason` implies it is an ad pause or use some kind of player state?

MelizzaP: we can assume its an ad if there is no `pauseReason`, in fact `adPause` is its own event.  But we won't be able to determine if scrolling the player out of view paused the ad or if the user paused it.

MelizzaP:
got some slack thumbs

PR #219, opened 2017-01-03
Liveblog responses

collin:


collin:
@kand I'm not sure what's going on with the `reactInstance` stuff, I try to checkout the files from master and nothing changes.

Otherwise this exists now...

elements/bulbs-liveblog/bulbs-liveblog-entry.js
```diff
@@ -16,6 +16,10 @@ class BulbsLiveblogEntry extends BulbsHTMLElement {
     if (this.liveblog && this.liveblog.handleEntryAttached) {
       this.liveblog.handleEntryAttached({ target: this });
     }
+
+    if (twttr) {
+      twttr.widgets.load();
+    }
```
kand: Perhaps we can note in the future a way to hook into an entry load finish so we can do this per site as needed.

collin: Yeah, we're not ready for it yet, but that is the dream of a `<bulbs-content>` or `<div is="bulbs-content">` wrapper.

It would be an element that does any content/widget initialization in the `attachedCallback`.

collin: Server would render any editor generated content in it:

```html
<div class="liveblog-entry-body" is="bulbs-content"></div>
<div class="liveblog-response-body" is="bulbs-content"></div>
```

PR #223, opened 2017-01-11
Custom elements v1

collin:


collin:
This would have to be given a bit more testing in test branches, but it's probably important that we get on the current spec.

PR #225, opened 2017-01-16
Add optional attribute to disable the ad refresh for one-off features

spra85:


elements/bulbs-dfp/bulbs-dfp.js
```diff
@@ -32,7 +32,7 @@ export default class BulbsDfp extends BulbsHTMLElement {
 
     let defaultRefreshInterval = 30000;
     this.adUnitData = this.adsManager.adUnits.units[this.dataset.adUnit];
-    if (!this.adUnitData.refreshDisabled) {
+    if (!this.adUnitData.refreshDisabled && !this.hasAttribute('no-refresh')) {
```
MelizzaP: should this be opposite? if it has `no-refresh` it should not set a refresh interval?

spra85: Should be right.  You only want to set up the refresh if it both does not have refresh disabled by the ad unit and does not have the `no-refresh` attribute.  There's test coverage for each scenario too.

MelizzaP: yuuuuuup, I read that weird.

PR #230, opened 2017-01-26
Bulbs pinned element edge case

MelizzaP:
# Bug
In certain situations, where an ad is not displayed on the rail and a user uses a shortcut to scroll all the way to the top of the page, the car escapes the bounds of the rail and pins itself to the top of the window, which is not the desired behavior.

# Test
This most reliably breaks when using adblock, but can be recreated without it. Use the `cmd` + `	▲` and `cmd` + `	▼` shortcut to scroll from the top of page to bottom or vice versa 
### Clickhole 
[Broken Pinned Element Link](http://www.clickhole.com/article/5-explorers-who-saw-americas-columbus-turned-back--5356)
[Fixed Pinned Element Test Link](http://bulbs-pinned-bump.test.clickhole.com/article/5-explorers-who-saw-americas-columbus-turned-back--5356)
### Avclub
[Broken Pinned Element Link](http://avclub.com/article/neal-brennan-why-white-dog-shit-funniest-kind-excr-248930)
[Fixed Pinned Element Test Link](http://bulbs-pinned-bump.test.avclub.com/article/neal-brennan-why-white-dog-shit-funniest-kind-excr-248930)


elements/bulbs-scroll-top/bulbs-scroll-top.js
```diff
@@ -0,0 +1,53 @@
+import {
+  registerElement,
+  BulbsHTMLElement,
+} from 'bulbs-elements/register';
+import invariant from 'invariant';
+
+export default class ScrollTop extends BulbsHTMLElement {
+  createdCallback () {
+    invariant(this.getAttribute('duration') < 0,
```
brycedorn: should this be `< 1` then? (0 not a pos int)

collin:
Oof. Well done here.

MelizzaP:
[bulbs-pinned-element sidebar behaves weird on scroll to top](https://trello.com/c/kHpnW8IO/62-bulbs-pinned-element-sidebar-behaves-weird-on-scroll-to-top)

elements/bulbs-scroll-top/bulbs-scroll-top.js
MelizzaP: good call, not going to be using scroll-top unfortunately :-(

MelizzaP:
@theonion/front-end this is ready for review

MelizzaP:
creating a ticket for clickhole, not enough time to fix that

kand:
I'm wondering if a possible fix here would be to get rid of the `position: fixed` on the car and instead  use `position: absolute` on the car and `position: relative` on the rail.

I think we might still have positioning problems when skipping up/down the page in large amounts, but at least the elements would be contained to the rail.

PR #231, opened 2017-01-26
create bulbs-pollzone

brycedorn:
moving from omni  ➡️  here, corresponds with https://github.com/theonion/omni/pull/765


brycedorn:
ready for :eyes: @theonion/front-end

elements/bulbs-pollzone-map/bulbs-pollzone-map.js
```diff
@@ -0,0 +1,121 @@
...
+    this.initMap();
+  }
+
+  setFills () {
+    this.fills = {
+      defaultFill: '#006B3A',
+    };
+
+    // add colors to keyed fills list
+    _.forEach([
```
collin: Is there any chance we could do this without the lodash dependency/find some way to use minimal imports (https://www.npmjs.com/package/lodash.topairs, etc.)

elements/bulbs-pollzone-map/bulbs-pollzone-map.js
```diff
@@ -0,0 +1,121 @@
...
+        })
+        .each((function (question) {
+          let votes = question[1];
+          let percent = Math.round(votes / stateResults.totalVotes * 100);
+
+          let i = this.questionsData[question[0]].sequence;
+          html += '<li class="result">' +
+            '<div class="bar" style="background-color:' + this.fillKey(i) + ';width: ' +
+            percent + '%;">' + '</div><div class="percent">' +
+            percent + '%</div></li>';
```
collin: And we could probably reduce this to a cleaner templatestring

```js
html += `
  <li class="result">
    <div class="bar" style="background-color: ${this.fillKey(i)}; width: ' ..... etc
`
```

elements/bulbs-pollzone-map/bulbs-pollzone-map.js
```diff
@@ -0,0 +1,121 @@
...
+import * as d3 from 'd3';
+import topojson from 'topojson';
+import Datamap from 'datamaps';
+
+import './bulbs-pollzone-map.scss';
+
+class BulbsPollzoneMap extends BulbsHTMLElement {
+  attachedCallback () {
+    let $this = $(this).find('.map');
+    let $data = $this.data('map-data');
```
collin: I think we could also do to replace these with dom equivalent:
```js
let map = this.querySelector('.map');
let data = map.getAttribute('data-map-data');

etc.
```

elements/bulbs-pollzone-map/bulbs-pollzone-map.js
brycedorn: I know, I tried:

```
import chain from 'lodash/chain'
import isEmpty from 'lodash/isEmpty'
import forEach from 'lodash/forEach'
import sortBy from 'lodash/orderBy'
```

but the chained functions were having issues (toPairs, sortBy)

MichaelButkovic:
I do Tuesday nights, so just right after Danniel works for me.

elements/bulbs-pollzone-map/bulbs-pollzone-map.js
collin: Yeah, I'll leave it to you. This one is mostly forward looking to when we maybe have to migrate some of these content-type specific elements for use in kinja and maybe don't want to bring all our dependencies with us.

brycedorn: it'll definitely happen at some point. I'd just rather have that be the task in itself, since there are (likely) other elements that will need the same refactor and it'd be easier to test if they're refactored at the same time

collin:
Makes sense :+1:

brycedorn:
kay @collin made those changes, redeploying to test branch on https://github.com/theonion/omni/pull/807 now

collin:
Tests are slurped by by a glob on `**/*.test.js`

collin:
Uses the weird webpack require.context:

https://github.com/theonion/bulbs-elements/blob/master/test/index.js#L12-L16

brycedorn:
right, makes sense. it was just stubbed differently than other tests (and would require some rewriting to adhere to bulbs-element structure)

elements/bulbs-pollzone-map/bulbs-pollzone-map.js
kand: I think we should be able to remove the lodash dep here, I'll write up examples in each instance to point you in the right direction. Let me know if you have any questions.

For this instance:

```
[
  // ... color list
].forEach(function (color, i) {
  this.fillKey(i + 1, color)
}, this);
```

elements/bulbs-pollzone-map/bulbs-pollzone-map.js
```diff
@@ -0,0 +1,123 @@
...
+      legendItems[i].style.backgroundColor = this.fillKey(i + 1);
+    }
+  }
+
+  showTooltip (geo, stateResults) {
+    let html = `<div class="hoverinfo"><div class="state-name">
+        ${geo.properties.name}
+      </div>`;
+
+    if (_.isEmpty(this.questionsData)) {
```
kand: Find a pollzone and check what type this field is. It looks like it's an array, so this should work as a replacement for that logic:

```
if (typeof this.questionsData === 'undefined' || this.questionsData.length < 1) {
  ...
```

elements/bulbs-pollzone-map/bulbs-pollzone-map.js
```diff
@@ -0,0 +1,123 @@
...
+          let votes = question[1];
+          let percent = Math.round(votes / stateResults.totalVotes * 100);
+
+          let i = this.questionsData[question[0]].sequence;
+          html += `<li class="result">
+            <div class="bar" style="background-color: ${this.fillKey(i)}; width: ${percent}%;"></div>
+            <div class="percent">${percent}</div>
+          </li>`;
+        }).bind(this))
+        .value();
```
kand: Replacement for `99` to `114`:

```
html += stateResults.votes.slice()
  .sort(function (question) {
    // not sure what the data format is here, so you'll have to find the equivalent, see sort documentation
  })
  .map(function (question) {
    let questionHtml = '';
    // build html markup, this will map each question to markup, then all of that will be appended to the html variable
    return questionHtml;
  });
```

[js `sort` function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)

elements/bulbs-pollzone-map/bulbs-pollzone-map.js
brycedorn: I believe it's a hash, but this should work the same

elements/bulbs-pollzone-map/bulbs-pollzone-map.js
brycedorn: :+1: I agree, collin said the same so I'll refactor with this

elements/bulbs-pollzone-map/bulbs-pollzone-map.js
kand: Ah okay, I think you can do `Object.keys(this.questionsData).length < 1` or some such check in that case.

brycedorn:
kay @kand :eye: pl0x

PR #233, opened 2017-02-02
setting currentArticle on new reading list item active item

shawncook:
Needed this to load disqus attributes for avclub reading list.

### DIFF
--- a/elements/bulbs-reading-list/lib/reading-list-articles.js
+++ b/elements/bulbs-reading-list/lib/reading-list-articles.js
@@ -34,8 +34,11 @@ export default class ReadingListArticles {
   }
 
   registerEvents () {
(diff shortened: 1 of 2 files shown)


elements/bulbs-reading-list/lib/reading-list-articles.js
```diff
@@ -35,6 +35,7 @@ export default class ReadingListArticles {
 
   registerEvents () {
     this.dispatcher.on('reading-list-item-clicked', this.handleListItemClicked.bind(this));
+    this.dispatcher.on('reading-list-item-url-changed', this.setCurrentArticle.bind(this));
```
brycedorn: just a convention I usually do, but you can set binds in the constructor of  rather than in the same line as calling them.

e.g. `this.setCurrentArticle = this.setCurrentArticle.bind(this)`, same for others. but just /2cents, not blocker

PR #234, opened 2017-02-02
add disableAds prop to revealed.js & test

benghaziboy:
@spra85 I think I got the negation correct for the line in question, but would like your blessing.
@kand @collin


benghaziboy:
[Modify the bulbs-video, revealed component to disable ad plugin if disabled for the video](https://trello.com/c/NEGrojBi/4-modify-the-bulbs-video-revealed-component-to-disable-ad-plugin-if-disabled-for-the-video)

PR #235, opened 2017-02-03
create bulbs-slick-slideshow

brycedorn:
corresponding pr in omni: https://github.com/theonion/omni/pull/767

should I rewrite the tests that were on omni? they're mostly testing basic functionality (which can be easily manually tested), but open to it since this is a pretty heavily used feature

### DIFF
--- a/elements/bulbs-pollzone-map/bulbs-pollzone-map.js
+++ b/elements/bulbs-pollzone-map/bulbs-pollzone-map.js
@@ -106,7 +106,7 @@ class BulbsPollzoneMap extends BulbsHTMLElement {
         let percent = Math.round(numVotes / totalVotes * 100);
 
         return `<li class='result'>
-          <div class='bar' style='background-color: ${this.fillKey(question['id'])}; width: ${percent}%;'></div>
+          <div class='bar' style='background-color: ${this.fillKey(question.id)}; width: ${percent}%;'></div>
           <div class='percent'>${percent}</div>
         </li>`;
       }, this).join('');
--- a/elements/bulbs-pollzone-map/bulbs-pollzone-map.js
+++ b/elements/bulbs-pollzone-map/bulbs-pollzone-map.js
@@ -106,7 +106,7 @@ class BulbsPollzoneMap extends BulbsHTMLElement {
         let percent = Math.round(numVotes / totalVotes * 100);
 
         return `<li class='result'>
-          <div class='bar' style='background-color: ${this.fillKey(question.id)}; width: ${percent}%;'></div>
+          <div class='bar' style='background-color: ${this.fillKey(question['id'])}; width: ${percent}%;'></div>
           <div class='percent'>${percent}</div>
         </li>`;
       }, this).join('');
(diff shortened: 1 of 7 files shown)


brycedorn:
@theonion/front-end ready for 👀 (and corresponding pr in desc)

elements/bulbs-slick-slideshow/slick.scss
```diff
@@ -0,0 +1,98 @@
+/* Slider */
+
+.slick-slider {
+    position: relative;
+    display: block;
+    box-sizing: border-box;
+    -webkit-touch-callout: none;
+    -webkit-user-select: none;
```
spra85: Can we make `user-select` a mixin?

elements/bulbs-slick-slideshow/slick.scss
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    &.dragging {
+        cursor: pointer;
+        cursor: hand;
+    }
+}
+.slick-slider .slick-track,
+.slick-slider .slick-list {
+    -webkit-transform: translate3d(0, 0, 0);
```
spra85: We should try to use a mixin for this as well

brycedorn: 👍 I didn't touch this file, it's just copied from the package itself.

reasoning was (aside from webpack not playing nice with importing css from a node module) that this keeps styles from changing if the package is changed.

brycedorn:
👍 to both comments above

elements/bulbs-slick-slideshow/slick.scss
collin: Same here, postcss/autoprefixer should cover us here.

elements/bulbs-slick-slideshow/slick.scss
collin: Ah yeah, I see a lot more is going on here than just prefixes. We should make a mixin and also reduce the number of prefixes if we don't need them.

elements/bulbs-slick-slideshow/bulbs-slick-slideshow.test.js
```diff
@@ -0,0 +1,94 @@
...
+      subject.appendChild(navElement);
+
+      attachSubject();
+      
+      sinon.spy(subject.slideshow, 'slick');
+    });
+
+    it('inits the jQuery slick carousel with the correct stuff', function () {
+      let { initialSlide } = subject;
+
```
brycedorn: stuck on this 👇 test 🤕 

I keep getting `AssertionError: expected slick to have been called with arguments`

spra85:
@brycedorn, take a look at the last few commits when you get in.  Fixed the test you were stuck on, will need an assist from you on testing this to make sure it still works for the slideshow content item.

brycedorn:
@kand re-👍?

elements/bulbs-pollzone-map/bulbs-pollzone-map.js
```diff
@@ -106,7 +106,7 @@ class BulbsPollzoneMap extends BulbsHTMLElement {
         let percent = Math.round(numVotes / totalVotes * 100);
 
         return `<li class='result'>
-          <div class='bar' style='background-color: ${this.fillKey(question.id)}; width: ${percent}%;'></div>
+          <div class='bar' style='background-color: ${this.fillKey(question['id'])}; width: ${percent}%;'></div>
```
kand: What issue does this change resolve?

brycedorn: this was just linting - actually removed from this PR (and merged earlier https://github.com/theonion/bulbs-elements/pull/238)

PR #238, opened 2017-02-07
pollzone lint fix

brycedorn:
travis wasn't catching this before but is now for some reason


brycedorn:
@theonion/front-end quick thumb, should fix lint

PR #242, opened 2017-02-08
bulbs-article-body

kand:
New element to apply things to an article body. Need this initially for appending dingbats, but it could be used for other things common to all articles.

### DIFF
--- a/elements/bulbs-article-body/README.md
+++ b/elements/bulbs-article-body/README.md
@@ -1,0 +1,15 @@
+# bulbs-article-body
+
+Provides some basic functionality every article body element needs. The element this is applied to must be a div.
+
+Example usage:
+
+```
+<div is="bulbs-article-body">
+  <!-- your article body content -->
+</div>
+```
+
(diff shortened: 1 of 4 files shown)


kand:

kand:
[use generalized reading list in ON and style](https://trello.com/c/dFCa1o5U/21-use-generalized-reading-list-in-on-and-style)

collin:
@spra85 I agree this hook could get out of hand.

I think the preference is a self-contained element. My guess is nearly all cases would be covered this way (eg. Embedding `<bulbs-video>` in an article body wouldn't need any support from `<bulbs-article-body>`). Something like _dingbat_ or  _inject a dfp slot in between paragraphs_ would be more the sort of thing that fits in this element.

spra85:
@collin, if it's not painful, can you put together a gist for how that might look with the 'inject a dfp slot' example and carry over the conversation to Slack?

kand:
@collin @spra85 are we good on the design?

PR #241, opened 2017-02-08
bulbs slideshow followups

brycedorn:
two things here:
 - 1. more precise selector (fixes reading list initialization issue)
 - 2. isolate key events (current bug on prod - left/right controls slideshows outside of viewport, this prevents that)

#### info on bug:

`init` calls `_.initADA();` which calls this 👇 

`_.$slides.add(_.$slideTrack.find('.slick-cloned')).attr({`

where `_.$slides` is apparently `null`. (and that's where it's throwing error)

`$slides` is initialized in a few places, notably in `buildOut` (L506 in slick.js): 

```
_.$slides =
            _.$slider
                .children( _.options.slide + ':not(.slick-cloned)')
```

so tried adding children. but that didn't work 😞

PR #243, opened 2017-02-09
bulbs slideshow followup v2

brycedorn:
cleaner commit than https://github.com/theonion/bulbs-elements/pull/241, didn't want to rebase

two things here:
 - 1. more precise selector (fixes reading list initialization issue for subsequent slideshows)
 - 2. isolate key events (current bug on prod - left/right controls slideshows outside of viewport, this prevents that)

### DIFF
--- a/elements/bulbs-slick-slideshow/bulbs-slick-slideshow.js
+++ b/elements/bulbs-slick-slideshow/bulbs-slick-slideshow.js
@@ -19,11 +19,7 @@ require('slick-carousel');
 /// 7. On l/r keydown, trigger slidechange.
 class BulbsSlickSlideshow extends BulbsHTMLElement {
   attachedCallback () {
-    this.slideshow = $('.slider');
-
-    if ($('.slider').length > 1) {
-      this.slideshow = $(this).find('.slider');
-    }
+    this.slideshow = $('.slider').last();
 
     // set variables
     this.initialSlide = 0;
--- a/elements/bulbs-slick-slideshow/bulbs-slick-slideshow.js
+++ b/elements/bulbs-slick-slideshow/bulbs-slick-slideshow.js
@@ -3,7 +3,8 @@ import {
   BulbsHTMLElement,
 } from 'bulbs-elements/register';
 import $ from 'jquery';
-import * as slick from 'slick-carousel'; // eslint-disable-line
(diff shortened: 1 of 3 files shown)


elements/bulbs-slick-slideshow/bulbs-slick-slideshow.js
```diff
@@ -19,6 +20,10 @@ class BulbsSlickSlideshow extends BulbsHTMLElement {
   attachedCallback () {
     this.slideshow = $('.slider');
 
+    if ($('.slider').length > 1) {
+      this.slideshow = $(this).find('.slider');
```
brycedorn: golden 🔑

two scenarios: 
 - there's only one slideshow on the page, so `length` is 1 and that is the slideshow we want to use
 - a slideshow has already been initialized, so use `this` to find the one currently in view

for some reason the latter doesn't work for the former case, don't ask me why but this is the only consistent fix I could get across test/browser

brycedorn:
@kand I went too deep into `slick`, still unsure what the bug is but this fixed it.

could possibly make a PR to slick so the second selector works every time, but not worth any more time on this.

elements/bulbs-slick-slideshow/bulbs-slick-slideshow.js
```diff
@@ -17,7 +19,7 @@ import './bulbs-slick-slideshow.scss';
 /// 7. On l/r keydown, trigger slidechange.
 class BulbsSlickSlideshow extends BulbsHTMLElement {
   attachedCallback () {
-    this.slideshow = $('.slider');
+    this.slideshow = $('.slider').last();
```
collin: is `.slider` inside the `<bulbs-slick-slideshow>` element?

Could this be scoped to

```js
this.slideshow = $(this).find('.slider').last();
```

brycedorn: yep, it really should be scoped to that. but for some reason when running this in tests this doesn't work (spent too much time digging/retrying different query selectors, no luck).

brycedorn: was never an issue in the browser, which is the weird part. but this still works, just somewhat slower since it queries the entire page.

PR #246, opened 2017-02-13
allow a more specific clickable area in reading list menus

kand:
Allows a more limited clickable area in a reading list menu item, addresses the issue where we want the feature type and campaign to be clickable for their own URLs apart from scrolling to a reading list article.


collin:
Could we achieve this by having the menu item ignore clicks that are on an `<a>` tag? Or is there something that prevents that?

kand:
An explicit area that is clickable makes for more sensible markup. With this change it will be obvious what areas in the markup correspond to scrolling an item as opposed to click logic that ignores some parts of the markup and is hidden in the the javascript.

collin:
👍

kand:

PR #247, opened 2017-02-13
some fixes for reading list menu precision

kand:
Fixes a few issues released in https://github.com/theonion/bulbs-elements/pull/246.


kand:

PR #249, opened 2017-02-14
Bulbs truncate element

shawncook:
Working PR for the article truncation

Test link:
http://avc-refresh-truncate-article.test.avclub.com/article/hear-rare-track-quiet-melancholy-los-campesinos-250252


elements/bulbs-truncate-element/bulbs-truncate-element.scss
```diff
@@ -0,0 +1,27 @@
+.bulbs-truncate-element-parent-active {
```
jmelvnsn: It works great. Maybe we can add a transition here? Would be a little nicer if we had a slight animation on the height of the active class or opacity attribute of the :after element. Maybe both?

jmelvnsn:
Other than small comment. lgtm

elements/bulbs-truncate-element/bulbs-truncate-element.scss
```diff
@@ -0,0 +1,27 @@
...
+    display: block;
+    height: 300px;
+    left: 0;
+    position: absolute;
+    right: 0;
+    top: auto;
+  }
+}
+
+.bulbs-truncate-element-button {
```
brycedorn: could add `outline: none` here to prevent that annoying blue border when clicking

elements/bulbs-truncate-element/bulbs-truncate-element.js
```diff
@@ -0,0 +1,38 @@
+import util from 'bulbs-elements/util';
+import {
+  registerElement,
+  BulbsHTMLElement,
+} from 'bulbs-elements/register';
+import './bulbs-truncate-element.scss';
+
+class BulbsTruncateElement extends BulbsHTMLElement {
```
collin: Nice element :)

elements/bulbs-truncate-element/bulbs-truncate-element.js
```diff
@@ -0,0 +1,38 @@
...
+
+  attachedCallback () {
+    this.addButton = this.addButton.bind(this);
+    this.openElement = this.openElement.bind(this);
+    this.addButton();
+    this.addEventListener('click', this.openElement);
+    $(this).prev().addClass('bulbs-truncate-element-parent-active');
```
collin: Good use of native dom methods in this element, another two apis we can use here

`this.previousElementSibling`

and `element.classList`

Together we can use them like:

```js
let previousEl = this.previousElementSibling;
if (previousEl) {
  previousEl.classList.add('bulbs-truncate-element-parent-active');
  previousEl.classList.remove('bulbs-truncate-element-parent-active');
}
```

Doing this now will help when we start migrating these components over to the `gawkermedia/kinja-components` repository which does not include jQuery as a depency.

elements/bulbs-truncate-element/bulbs-truncate-element.scss
shawncook: Leaving off the animation because it's not quick and easy to animate between `max-height:1000px` and `max-height:auto`. We can come back to this after QA has a look at the functionality.

PR #251, opened 2017-02-15
Add Video Ad Config dimension

benghaziboy:


benghaziboy:
released via
https://github.com/theonion/bulbs-elements/releases/tag/11.11.3

PR #257, opened 2017-02-22
pass & test

benghaziboy:
@spra85 @collin

### DIFF
--- a/elements/bulbs-video/elements/rail-player/components/root.js
+++ b/elements/bulbs-video/elements/rail-player/components/root.js
@@ -57,5 +57,6 @@ Root.propTypes = {
   channel: PropTypes.string,
   recircUrl: PropTypes.string.isRequired,
   targetCampaignId: PropTypes.string,
+  targetSpecialCoverage: PropTypes.string,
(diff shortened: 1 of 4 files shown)


benghaziboy:
y

benghaziboy:
Yes. This mirrors the behavior of the `targetCampaignId` prop in the `right-rail` so I had to
- Add to the `rail-player` component
- Add to the `rail-player/root` component
- Props are then automatically passed into Revealed and handled as they are other in other contexts.

spra85:
LGTM @benghaziboy.  Guessing then we'll need an additional change in the `omni` repo that passes the new attr in?

benghaziboy:
Yes. Doing a temp release to make that change atm.

benghaziboy:
[Pass `targetSpecialCoverage` through the elements `rail-player.js` component & the `rail-player/root.js` component to `revealed.js](https://trello.com/c/oNaQWmel/1257-pass-targetspecialcoverage-through-the-elements-rail-player-js-component-the-rail-player-root-js-component-to-revealed-js)

PR #259, opened 2017-02-22
removing event listener for now

brycedorn:
looks like this was autoplaying first video regardless of the second video that is selected. can revisit afterward for followup fix, this (should) just disable annoying audio.

corresponding omni pr: https://github.com/theonion/omni/pull/899

### DIFF
--- a/elements/bulbs-video/elements/video-carousel.js
+++ b/elements/bulbs-video/elements/video-carousel.js
@@ -57,7 +57,10 @@ class BulbsVideoCarousel extends BulbsHTMLElement {
       '<bulbs-video-carousel> MUST contain a <bulbs-carousel>'
     );
 
-    this.videoPlayer.addEventListener('jw-beforePlay', this.firstPlay = this.firstPlay.bind(this), true);
+    // Removing for now, causing issue with videos autoplaying in background 
+    //  - bryce, 2/22/17
+    // this.videoPlayer.addEventListener('jw-beforePlay', this.firstPlay = this.firstPlay.bind(this), true);
+
     this.videoPlayer.addEventListener('jw-complete', this.playerEnded = this.playerEnded.bind(this), true);
     this.carousel.addEventListener('click', this.handleClick = this.handleClick.bind(this));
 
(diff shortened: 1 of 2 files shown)


elements/bulbs-video/elements/video-carousel.js
```diff
@@ -57,7 +57,10 @@ class BulbsVideoCarousel extends BulbsHTMLElement {
       '<bulbs-video-carousel> MUST contain a <bulbs-carousel>'
     );
 
-    this.videoPlayer.addEventListener('jw-beforePlay', this.firstPlay = this.firstPlay.bind(this), true);
+    // Removing for now, causing issue with videos autoplaying in background 
+    //  - bryce, 2/22/17
+    // this.videoPlayer.addEventListener('jw-beforePlay', this.firstPlay = this.firstPlay.bind(this), true);
```
spra85: @collin can you weigh in here?

brycedorn:
FWIW, is a bug on ON homepage as well.

elements/bulbs-video/elements/video-carousel.js
collin: I think this is listening to the first play event to update the styling on the carousel items.

elements/bulbs-video/elements/video-carousel.js
```diff
@@ -68,8 +71,9 @@ class BulbsVideoCarousel extends BulbsHTMLElement {
 
   firstPlay () {
     let items = this.querySelectorAll('bulbs-carousel-item');
+    let nowPlayingItems = this.querySelectorAll('[now-playing]');
 
-    if (items.length > 0) {
+    if (nowPlayingItems.length > 0 && items.length > 0) {
```
collin: Which calls through to this. The apply state function is doing two things:

1: updating the styles on the carousel item to highlight the current playing item.
2: sets `src` attribute of the video player. Changing the `src` attribute of a `<bulbs-video>` tears down the jwplayer instance and creates a new one.

elements/bulbs-video/elements/video-carousel.js
collin: Losing that styling for a moment is probably better than playing two stacked videos as a bug.

elements/bulbs-video/elements/video-carousel.js
brycedorn: but the `selectItem` call below is currently forcing `items[0]` to play, regardless of which one is selected. the highlight styles don't seem to be related to `firstPlay`, as this is only called once (and then the eventlistener is removed), so I'm not really sure what the intended functionality of this is

brycedorn:
merging with plans to followup with better soln soon.

PR #261, opened 2017-02-24
Switch to Akamai CDN static assets

mparent61:
[The Asset In Question](http://assets2.onionstatic.com/onion/static/images/ad-block-check.jpg)


mparent61:
[Switch all sites to use Akamai for static assets](https://trello.com/c/r1HrOPhK/80-switch-all-sites-to-use-akamai-for-static-assets)

PR #262, opened 2017-02-24
bulbs-lightbox element

jmelvnsn:
@theonion/front-end 

Lightbox element to replace old Onion implementation to enlarge article images.

https://github.com/theonion/omni/pull/915


elements/bulbs-lightbox/bulbs-lightbox.js
```diff
@@ -0,0 +1,36 @@
...
+          this.classList.add('active');
+        }
+      }
+
+      // Close overlay with Esc key
+      $(document).keyup(function(e) {
+        if (e.keyCode == 27) overlay.classList.remove('active');
+      });
+
+      this.addEventListener('click', toggleOverlay);
```
collin: I think we ought to translate these event listeners to be more consistent with the bulbs-elements house style.

Something like:

```js
document.addEventListener('keyup', this.handleKeyUp.bind(this));
document.addEventListener('click', this.toggleOverlay.bind(this));
```

with `toggleOverlay` and `handleKeyUp` defined as methods on element class.

elements/bulbs-lightbox/bulbs-lightbox.scss
```diff
@@ -0,0 +1,36 @@
...
+    height: 100vh;
+    justify-content: center;
+    left: 0;
+    opacity: 0;
+    position: fixed;
+    right: 0;
+    top: 0;
+    transform: scale(0.9);
+    transition: transform 0.2s, opacity 0.2s, visibility 0s 0.2s;
+    visibility: hidden;
```
brycedorn: looks like this [won't load the img asynchronously](http://stackoverflow.com/questions/578283/will-an-image-load-if-it-has-its-css-set-so-that-visibility-is-hidden), is that intended?

I made a library kinda like this a little while ago - will help with page load if just the preview is loaded (and the larger version onclick)

elements/bulbs-lightbox/bulbs-lightbox-examples.js
```diff
@@ -0,0 +1,27 @@
+import './bulbs-lightbox.scss';
+import './bulbs-lightbox-test.scss';
+
+let examples = {
+  element: 'bulbs-lightbox',
+  examples: {
+    'Bulbs LightBox': {
+      render: () => {
+        return `
+          <bulbs-lightbox>
```
spra85: @jmelvnsn opted to make this a semantic component as opposed to `is="bulbs-lightbox"`

jmelvnsn: yeah. little easier to read/comprehend in the document source

PR #266, opened 2017-03-02
git-cher-index

collin:
Something like this...


collin:
@shawncook am I good to delete this branch now?

shawncook:
i deleted!

PR #267, opened 2017-03-03
debounce events around animation frames

collin:
Extracting this out as a general utility for reading list concerns.

```js
import { debouncePerFrame } from 'bulbs-elements/util';
let debouncer = debouncePerFrame();

debouncer(() => { console.log('a') });
debouncer(() => { console.log('b') });

// only 'b' will log

// pageStartDebouncer will only execute the LAST callback passed
// to it per animation frame
//
// You can create multiple debouncers and they will not interract with each other:
let debouncerA = debouncePerFrame();
let debouncerB = debouncePerFrame();

debouncerA(() => { console.log('a') })
debouncerB(() => { console.log('b') })

// `a` and `b` will log
```


collin:
@theonion/front-end review plz

collin:
This appears to work the same as it currently does on production:

http://www.theonion.com/interactive/obama
http://bulbs-page-update.test.theonion.com/interactive/obama

PR #268, opened 2017-03-06
Add a new value for type to be used for possible preroll exclusion

spra85:
This simply adds a new possible value for the`type` custom param key/value, to pass in a blanket `sponsored` type in instances where a stand-alone video might be sponsored (defined herein as the presence of the `targetCampaignId`).  It will make it easier to globally exclude any sponsored video from affected line items on the ops side.

Will be a `minor` release.

### DIFF
--- a/elements/bulbs-video/components/revealed.test.js
+++ b/elements/bulbs-video/components/revealed.test.js
@@ -641,6 +641,7 @@ describe('<bulbs-video> <Revealed>', () => {
           let parsed = url.parse(vastUrl, true);
           let cust_params = querystring.parse(parsed.query.cust_params, '&');
           expect(cust_params.dfp_campaign_id).to.equal('1');
+          expect(cust_params.type).to.equal('sponsored');
         });
       });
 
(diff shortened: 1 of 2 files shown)


elements/bulbs-video/components/revealed.test.js
```diff
@@ -641,6 +641,7 @@ describe('<bulbs-video> <Revealed>', () => {
           let parsed = url.parse(vastUrl, true);
           let cust_params = querystring.parse(parsed.query.cust_params, '&');
           expect(cust_params.dfp_campaign_id).to.equal('1');
+          expect(cust_params.type).to.equal('sponsored');
```
collin: do we have a case covering type='special_coverage'?

spra85: Yeah, @collin -> https://github.com/theonion/bulbs-elements/blob/7906a3a65c861c2e1c5c95a0cc459b564354a465/elements/bulbs-video/components/revealed.test.js#L662

PR #271, opened 2017-03-09
unified analytics

brycedorn:
### what this does:

 - ~~rm references to `avclubAnalytics`, `onionan`, `clickholean`~~
   - ~~replace with `analyticsManager`~~

### todo
 - create new analytics manager for slideshow/clickventure


lib/bulbs-elements/util/get-analytics-manager.js
```diff
@@ -9,22 +9,16 @@ FIXME: videohub-player depends on there being an instance of our analytics manag
...
            lives.
 */
-/* global avclubAnalytics, onionan, clickholean, starwipe */
+/* global starwipe */
 
 export default function () {
-  let analyticsManager;
-  if (window.avclubAnalytics) {
-    analyticsManager = avclubAnalytics;
+  if (window.starwipe) {
```
brycedorn: pr [here](https://github.com/theonion/starwipe/pull/291) to consolidate this further

collin:
Exciting change-sets here

elements/bulbs-slick-slideshow/bulbs-slick-slideshow.js
```diff
@@ -109,10 +113,15 @@ class BulbsSlickSlideshow extends BulbsHTMLElement {
...
   slideshowChanged (event, slickObject, currentSlide) {
     window.location.hash = currentSlide;
-    window.onionan.trackPageView(true);
+    this.trackPageView();
     this.enableDisableNav(this.slides, currentSlide);
   }
 
+  trackPageView () {
+    this.analyticsManager.setDimension('navigation_method','slideshow');
+    this.analyticsManager.trackPageView(false);
```
brycedorn: aligning this with clickventure - this `false` param pertains to `freshPage` param which is called with `sendQuantcastPixel` & `sendComscorePixel`

should keep this consistent, since if clickventure isn't considering another frame to be a new page a slideshow shouldn't either.

elements/bulbs-clickventure/clickventure.js
```diff
@@ -112,6 +112,7 @@ export default class Clickventure {
 
         clickventure.nodeClickCount++;
         clickventure.gotoNodeId(targetNode, transitionName);
+        this.analyticsManager.setDimension('navigation_method','clickventure');
```
brycedorn: https://github.com/theonion/bulbs-public-analytics-manager/pull/13

lib/bulbs-elements/util/get-analytics-manager.js
```diff
@@ -9,22 +9,10 @@ FIXME: videohub-player depends on there being an instance of our analytics manag
...
-  }
-  else if (window.clickholean) {
-    analyticsManager = clickholean;
-  }
-  else if (window.starwipe) {
-    analyticsManager = starwipe.analyticsManager;
-  }
-  return analyticsManager;
+  // Global default
+  return window.analyticsManager;
```
brycedorn: can get rid of this file now - as followup though to not make this pr massive

brycedorn:
having issues w/ child not hearing global async `eventEmitter` on content load - it seems to be firing but the listener doesn't pick it up

@collin does this look right? or should I just scrap this and pass the `gaPrefix` like a normal data var (and reinitialize like here https://github.com/theonion/bulbs-elements/blob/master/lib/bulbs-elements/util/prep-ga-event-tracker.js)?

elements/bulbs-slick-slideshow/bulbs-slick-slideshow.js
```diff
@@ -29,6 +29,9 @@ class BulbsSlickSlideshow extends BulbsHTMLElement {
     this.navPrev = this.navLinks.find('.slider-prev');
     this.restart = this.slideshow.find('a.restart');
 
+    // listen for GA
+    this.addEventListener('ga-analytics-manager-initialized', this.setAnalyticsManager.bind(this));
```
collin: @brycedorn is this the event you're talking about?

elements/bulbs-slick-slideshow/bulbs-slick-slideshow.js
spra85: We may want to also add some protection into this method, to at least check for the presence of `this.analyticsManager` and log an error if it's not there

elements/bulbs-slick-slideshow/bulbs-slick-slideshow.js
brycedorn: yep

brycedorn:
seems to be working now; releasing tagged version

brycedorn:
@jmelvnsn hello it is me

brycedorn:
would like re-approval since so many subsequent changes @jmelvnsn @collin @spra85

spra85:
@brycedorn, can you provide relevant test links in an Omni PR

elements/bulbs-slideshow/bulbs-slideshow.js
```diff
@@ -103,9 +109,18 @@ class BulbsSlideshow extends BulbsHTMLElement {
 
   afterSlideTransition () {
     this.render();
+    this.sendPageView.bind(this).call();
```
spra85: I wouldn't have guessed you'd need to bind `this` anywhere for these calls because this function itself is not called as an event listener.  Can you describe where you were running into context issues?

lib/bulbs-elements/util/prep-reading-list-analytics.js
```diff
@@ -0,0 +1,19 @@
...
+    this.href = parent.dataset.href;
+    this.title = parent.dataset.title;
+  }
+
+  let dimensionObject = {
+    dimension12: dimension12,
+  };
+
+  this.analyticsManager = getAnalyticsManager();
+  this.analyticsWrapper = prepGaWrapper(this.parentAnalyticsDimensions, dimensionObject);
```
spra85: I'm not crazy about the pattern to have this external function modifying `this` as applied at time of method invocation.  I'd prefer to use ES6 destructuring and return multiple values and leave it up to the integrating object to define values on `this`.  Thoughts @collin ?

spra85: If this pattern were to be used in more places than this, you'd have to look all over the place to find property definitions for objects

elements/bulbs-slideshow/bulbs-slideshow.js
```diff
@@ -103,9 +109,18 @@ class BulbsSlideshow extends BulbsHTMLElement {
...
     this.navigateToNextSlide();
   }
 
+  sendPageView () {
+    this.analyticsManager.trackPageView(
```
spra85: This `sendPageView` method is duplicated in enough places where it makes sense to wonder if we should bubble this up to a util as well, and pass in the relevant values?

elements/bulbs-slick-slideshow/bulbs-slick-slideshow.js
```diff
@@ -29,6 +24,8 @@ class BulbsSlickSlideshow extends BulbsHTMLElement {
     this.navPrev = this.navLinks.find('.slider-prev');
     this.restart = this.slideshow.find('a.restart');
 
+    prepReadingListAnalytics.apply(this, [this.slideshow, 'slideshow']);
```
spra85: See comment below on this pattern.

lib/bulbs-elements/util/prep-reading-list-analytics.js
collin: I'm with Chris here. 

I think also in general calling a utility function with a bound `this` rather than having a regular named argument is a little confusing.

`function foo (arg2, arg3)` vs `function foo (arg1, arg2, arg3)`

elements/bulbs-slideshow/bulbs-slideshow.js
brycedorn: this would face the same issue as above - either binding context or passing in 4 values

spra85: Comment about duplication aside, I don't see why you need to `.bind(this)` for calls to `sendPageView`

lib/bulbs-elements/util/prep-reading-list-analytics.js
brycedorn: this is true, was experimenting with this pattern - since most of the logic in `prepReadingListAnalytics` is initializing/setting values on `this` it seemed worth extracting out. 

would it be more clear if the method had `setValues` in the name? or should this be something that should just be copied in each element?

elements/bulbs-slideshow/bulbs-slideshow.js
brycedorn: `sendPageView` doesn't have access to `analyticsManager` or `analyticsWrapper`, etc IIRC

elements/bulbs-slideshow/bulbs-slideshow.js
brycedorn: `sendPageView` doesn't have access to `analyticsManager` or `analyticsWrapper`, etc IIRC

lib/bulbs-elements/util/prep-reading-list-analytics.js
collin: I think chris makes a good point about being clear in the class when properties are getting set.

@spra85 @brycedorn what would you think about about a pattern like this?

```js
Object.assign(this, makeReadingListAnalyticsProps(this, element, 'element'))
```

Otherwise the more explicit would be:

```js
const readingListProps = makeReadingListAnalyticsProps(this, element, 'element'))
this.foo = readingListProps.foo;
this.bar = readingListProps.bar;
```

brycedorn: I'm 👌  with the former - basically the same functionality but the assignment happens in the element itself.

spra85: And I'm 👍 with the latter, because I think it's far more explicit and readable

elements/bulbs-slick-slideshow/bulbs-slick-slideshow.js
```diff
@@ -109,17 +110,26 @@ class BulbsSlickSlideshow extends BulbsHTMLElement {
...
+    );
+  }
+
   bindContext () {
+    this.bodyKeyDown = this.bodyKeyDown.bind(this);
     this.navNextClicked = this.navNextClicked.bind(this);
     this.navPrevClicked = this.navPrevClicked.bind(this);
-    this.bodyKeyDown = this.bodyKeyDown.bind(this);
+    this.restartShow = this.restartShow.bind(this);
+    this.sendPageView = this.sendPageView.bind(this);
```
spra85: Can we confirm we'd still need this `bind`

brycedorn: you're right, shouldn't be necessary

brycedorn:
@spra85 fixed that issue you were seeing, should be 👍 for final 👁

elements/bulbs-slideshow/bulbs-slideshow.js
```diff
@@ -103,9 +107,31 @@ class BulbsSlideshow extends BulbsHTMLElement {
...
+    this.title = readingListProps.title;
+  }
+
+  sendPageView () {
+    if (!this.analyticsManager) {
+      this.setupGA();
+    }
+
+    this.analyticsManager.trackPageView(
+      this.href,
```
spra85: Should be `false`, the argument is `freshPage` which expects a boolean

elements/bulbs-slick-slideshow/bulbs-slick-slideshow.js
```diff
@@ -109,17 +104,39 @@ class BulbsSlickSlideshow extends BulbsHTMLElement {
...
+    this.title = readingListProps.title;
+  }
+
+  sendPageView () {
+    if (!this.analyticsManager) {
+      this.setupGA();
+    }
+
+    this.analyticsManager.trackPageView(
+      this.href,
```
spra85: Comment below, `href` should be `false`

elements/bulbs-clickventure/clickventure.js
```diff
@@ -228,4 +236,12 @@ export default class Clickventure {
       this.adsManager.reloadAds(this.sideAd);
     }
   }
+
+  sendPageView () {
+    this.analyticsManager.trackPageView(
+      this.href,
```
spra85: Also, re other comments, `this.href` argument should be the boolean `false` here

elements/bulbs-slideshow/bulbs-slideshow.js
```diff
@@ -103,9 +107,31 @@ class BulbsSlideshow extends BulbsHTMLElement {
...
+  setupGA () {
+    const readingListProps = prepReadingListAnalytics(this.$slideshow, { dimension12: 'slideshow' });
+    const { analyticsManager, analyticsWrapper, title } = readingListProps;
+
+    this.analyticsManager = analyticsManager;
+    this.analyticsWrapper = analyticsWrapper;
+    this.title = title;
+  }
+
+  sendPageView () {
```
brycedorn: so @spra85 if you could give this another pair of 👀, I dug into this and was able to confirm the following:
 - this event (`sendPageView`) _always_ fires when clicking next/previous
   - subsequently, `trackPageView` always fires
 - the only `ga('send','pageview')` events that the GA debugger picks up are those _after_ the 1st time 'next' is clicked
    - there's no code that specifies that the event shouldn't fire on the first click; everything 
 needed exists. 
    - and the code is (seemingly) identical to clickventure's (which fires on 1st 'next')

PR #272, opened 2017-03-10
Flyout scroll lock

collin:


shawncook:
bitmoji

pending master merge :D

collin:
talked to @kand, different issue that we're okay with over the weekend

PR #273, opened 2017-03-13
pulling out analyicsManager change

brycedorn:
going to roll this out separately to start getting thing out - this just updates the global var to be `analyticsManager` 

was a part of https://github.com/theonion/bulbs-elements/pull/271 but will continue work on that for passing analytics manager through to track events correctly


brycedorn:
eh let's just release it all at once, fixed other issue

PR #276, opened 2017-03-14
rm (deprecated) ClickVenture event

brycedorn:
screen shot 2017-03-14 at 2 20 40 pm

kept getting this^ on prod, doesn't look like there's a `collapsynav` element anymore. also removed event exclusively tied to this

# Test

http://ch-cv-error-fix.test.clickhole.com/clickventure/welcome-defensive-driving-chumps-youre-mine-next-6-5626


brycedorn:
PSA @theonion/front-end

brycedorn:
@collin I just saw it in console when clicking through one - can't find any reference to the element it's trying to collapse

PR #278, opened 2017-03-15
Revert "Reading list Refactor"

brycedorn:
may not need - just opening in case


brycedorn:
@theonion/front-end PSA

PR #279, opened 2017-03-15
Reading list Refactor v2

brycedorn:
Reverts theonion/bulbs-elements#278 (reverts revert)


elements/bulbs-reading-list/components/bulbs-reading-list-item.js
```diff
@@ -3,13 +3,149 @@ import {
...
+
+let pageStartDebouncer = debouncePerFrame();
 
 class BulbsReadingListItem extends BulbsHTMLElement {
   attachedCallback () {
-    invariant(this.dataset.id, '<bulbs-reading-list-menu> requires a "data-id" attribute');
-    invariant(this.dataset.href, '<bulbs-reading-list-menu> requires a "data-href" attribute');
-    invariant(this.dataset.partialUrl, '<bulbs-reading-list-menu> requires a "data-partial-url" attribute');
-    invariant(this.dataset.title, '<bulbs-reading-list-menu> requires a "data-title" attribute');
+    const markdownText = '<bulbs-reading-list-item> requires attribute: ';
```
brycedorn: thought is to add a `is-loaded` attribute here so the template [here](https://github.com/theonion/omni/blob/master/avclub/avclub/templates/core/detail/_base.html#L64) doesn't reinitialize it

lib/bulbs-elements/util/in-view-monitor.js
```diff
@@ -128,7 +138,9 @@ function handleDisplayMutation () {
       });
 
       Object.keys(monitoredItems).forEach((key) => {
-        maybeEmitEvent(monitoredItems[key]);
+        if(monitoredItems[key]) {
```
spra85: Is this `if` conditional actually needed?  If we're already iterating over the keys of the object, they should always be there, no?

MelizzaP: yeah, I was running into errors around `monitoredItems[key]` being undefined and breaking the scroll/resize events

PR #280, opened 2017-03-20
rail player style tweaks

brycedorn:


collin:
Confirmed that this is an issue with stock jwplayer css when the player width is small and the aspect ratio of the video container is constrained to 16x9.

PR #281, opened 2017-03-20
Multi entry analytics setup

jmelvnsn:
@spra85 @brycedorn 

https://github.com/theonion/omni/pull/1032

http://multi-entry-analytics.test.theonion.com/live/president-donald-trump-first-100-days-55086

### DIFF
--- a/elements/bulbs-page/bulbs-page.js
+++ b/elements/bulbs-page/bulbs-page.js
@@ -91,7 +91,7 @@ export default class BulbsPage extends BulbsHTMLElement {
   sendAnalyticsEvent () {
     this.gaTrackerWrapper(
       'send', 'event',
-      'multi_entry',
+      'reading_list',
       'scroll_view',
       this.getAttribute('pushstate-url')
     )
--- a/elements/bulbs-page/bulbs-page.js
+++ b/elements/bulbs-page/bulbs-page.js
@@ -94,7 +94,7 @@ export default class BulbsPage extends BulbsHTMLElement {
       'reading_list',
       'scroll_view',
       this.getAttribute('pushstate-url')
-    )
+    );
   }
 
 }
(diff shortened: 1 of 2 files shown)


elements/bulbs-page/bulbs-page.js
```diff
@@ -47,10 +81,22 @@ export default class BulbsPage extends BulbsHTMLElement {
...
+        this.gaTrackerWrapper
       );
+      this.sendAnalyticsEvent();
     });
   }
+
+  sendAnalyticsEvent () {
+    this.gaTrackerWrapper(
+      'send', 'event',
+      'multi_entry',
```
jmelvnsn: @spra85 is this how we want to refer to multi entry in GA?

jmelvnsn: @spra85 or do we need to replace all instances of liveblog with multi_entry?

elements/bulbs-page/bulbs-page.js
```diff
@@ -28,6 +29,35 @@ export default class BulbsPage extends BulbsHTMLElement {
     InViewMonitor.remove(this);
   }
 
+  get dimensions () {
```
brycedorn: `getDimensions` I think

brycedorn: if you want to pull this out into `'bulbs-elements/util'`, I was planning on doing that. seems this will be in a few places just to keep things dry

elements/bulbs-page/bulbs-page.js
spra85: @jmelvnsn, I think we just want to refer to them in GA this way

elements/bulbs-page/bulbs-page.js
spra85: Agree with @brycedorn here, the two of you should work together on bubbling this up to `util` so it can be used in all 4 places: 

- `bulbs-page`
- `reading-list-article`
- Slick Slideshow
- Clickventure

jmelvnsn:
@spra85 that is correct

brycedorn:
@spra85 👍 on this? I can ship and then make that util in my followup

PR #283, opened 2017-03-22
rail player volume fix

brycedorn:
the volume slider is still being wonky


brycedorn:
this is just a followup to https://github.com/theonion/bulbs-elements/pull/280, going to merge & patch

PR #285, opened 2017-03-22
mute on unmount to prevent audio from continuing to play

kand:
Prevents a removed video's audio from playing in the case that it hasn't been cleaned up by jwplayer's `remove` function. A hack since it seems the video is still around in memory, but solves the problem of audio playing.


kand:
[audio doesn't pause when you click on multiple videos within the AVC homepage player carousel, they overlay on top of one another](https://trello.com/c/4DIrkAJN/1320-audio-doesn-t-pause-when-you-click-on-multiple-videos-within-the-avc-homepage-player-carousel-they-overlay-on-top-of-one-another)

PR #286, opened 2017-03-23
Bulbs load more gif

MelizzaP:
# update
Working on the avclub's newswire reading list, I noticed we didn't have anything notifing users that a fetch is pending. This pr adds a loading gif while fetch pending in bulbs-load-more button


MelizzaP:
@theonion/front-end

PR #289, opened 2017-03-28
put campaign-display in video carousel

kand:
Put `campaign-display` element in video carousel.

### Release Type: _major_
Video sponsorship data is now pulled from a different location. Potentially breaking, but should not be based on the way authors are uploading and sponsoring videos.

### Test Link
- [AV](http://vid-carousel-summary-campaign.test.avclub.com/)

### Added
1. `bulbs-video` `VideoSummaryView` now includes a `campaign-display` if given video is sponsored through tunic.

### Updated
1. `bulbs-elements` will check that an element has a react instance before rendering when an attribute has changed. This fixes a problem where element constructors are called before they have access to all of their attributes which caused React to throw prop type validation errors before the element has access to its entire `props` object.

### Removed
1. `campaign-display` check for placement attribute removed in favor of React's out of the box prop types validation for this property.

### DIFF
--- a/elements/bulbs-video/elements/meta/components/root.js
+++ b/elements/bulbs-video/elements/meta/components/root.js
@@ -74,7 +74,6 @@ VideoMetaRoot.propTypes = {
   campaignPlacement: PropTypes.string,
   campaignPreamble: PropTypes.string,
   campaignTrackAction: PropTypes.string,
-  campaignUrl: PropTypes.string,
   disableLink: PropTypes.bool,
   mobileTitle: PropTypes.string,
   relativeSeriesLinkPrefix: PropTypes.string,
--- a/elements/bulbs-video/elements/meta/components/root.js
+++ b/elements/bulbs-video/elements/meta/components/root.js
@@ -37,13 +37,24 @@ export default function VideoMetaRoot (props) {
     seriesElement = seriesNameElement;
   }
 
+  let campaign;
+  if (props.video.tunic_campaign_url) {
+    campaign = (
+      <VideoMetaCampaign
+          campaignPlacement={props.campaignPlacement}
+          campaignPreamble={props.campaignPreamble}
+          campaignTrackAction={props.campaignTrackAction}
+          campaignUrl={props.video.tunic_campaign_url} />
+    );
+  }
+
   return (
     <div className='bulbs-video-meta'>
       <div className='bulbs-video-meta-copy'>
 
         { seriesElement }
 
-        { props.campaignUrl ? <VideoMetaCampaign {...props}/> : null }
+        {campaign}
 
         <bulbs-ellipsize class='bulbs-video-meta-title' line-count="3">
           {props.video.title}
--- a/elements/bulbs-video/elements/summary.js
+++ b/elements/bulbs-video/elements/summary.js
@@ -24,6 +24,19 @@ export function VideoSummaryView (props) {
     );
   }
 
+  let campaign;
+  if (video.tunic_campaign_url) {
+    campaign = (
+      <campaign-display
+        data-track-action={props.campaignTrackAction}
+        placement={props.campaignPlacement}
+        preamble-text={props.campaignPreamble}
+        src={video.tunic_campaign_url}
+        name-only
+      />
+    );
+  }
+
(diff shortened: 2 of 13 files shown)


kand:
[Show sponsorship within Homepage Video player carousel](https://trello.com/c/ypzSe0Ay/1293-show-sponsorship-within-homepage-video-player-carousel)

elements/bulbs-video/elements/summary.js
```diff
@@ -24,6 +24,19 @@ export function VideoSummaryView (props) {
...
 
+  let campaign;
+  if (video.tunic_campaign_url) {
+    campaign = (
+      <campaign-display
+        data-track-action={props.campaignTrackAction}
+        placement={props.campaignPlacement}
+        preamble-text={props.campaignPreamble}
+        src={video.tunic_campaign_url}
+        name-only
```
collin: Looks okay?

Does this world if you 'camelCase' the attributes?

kand: I've tried that as well, in every case it only takes the first attribute defined on the element.

elements/bulbs-video/elements/meta/components/root.js
```diff
@@ -37,13 +37,24 @@ export default function VideoMetaRoot (props) {
...
+  }
+
   return (
     <div className='bulbs-video-meta'>
       <div className='bulbs-video-meta-copy'>
 
         { seriesElement }
 
-        { props.campaignUrl ? <VideoMetaCampaign {...props}/> : null }
+        {campaign}
```
kand: @collin do you see any issues with using the sponsor from the video object itself instead of the url which is passed in from the site's CMS.

From what I've gathered AV is putting sponsorships in both places so these fields represent the same thing essentially. With onion studios though the video will show sponsorship across all usages instead of only on the site.

kand: To follow up, at this point it's fastest to leave the code as I have it changed here. Both fields should pull from the same place, let me know if there's an outstanding issue with doing it this way and I can work to remove these changes.

elements/bulbs-video/elements/summary.test.js
```diff
@@ -71,8 +71,14 @@ describe('<bulbs-video-summary>', () => {
     it('renders <VideoSummaryView>', () => {
       subject = shallow(<BulbsVideoSummary nowPlaying=''/>);
       subject.setState({ video });
+
       expect(subject.equals(
-        <VideoSummaryView video={video} nowPlaying={true}/>
+        <VideoSummaryView
```
brycedorn: if you're manually setting these to `null` could you just have `defaultProps` set it to `null`?

elements/bulbs-video/elements/summary.js
```diff
@@ -24,6 +24,19 @@ export function VideoSummaryView (props) {
     );
   }
 
+  let campaign;
+  if (video.tunic_campaign_url) {
+    campaign = (
+      <campaign-display
+          data-track-action={props.campaignTrackAction}   // eslint-disable-line react/prop-types
```
brycedorn: why the eslint-disable? does it want you to do `const { campaignTrackAction, campaignPlacement, campaignPreamble } = props`? (cleaner imo)

elements/bulbs-video/elements/meta/components/root.js
collin: I can't speak to which url/sponsor we need to use here in terms of supported workflow and process..

elements/bulbs-video/elements/summary.test.js
kand: Took another look at what I did here, turned out to be unnecessary, removed this change entirely.

elements/bulbs-video/elements/summary.js
kand: Turned out also didn't need this change. ESlint wanted these to be defined in `propTypes`, but there are two `propTypes` in this file which I didn't notice, so I thought this was complaining because these were undefined but that was incorrect.

kand:
@brycedorn you wanna re-review and either 👍 or 👎 ?

PR #291, opened 2017-03-29
fix more bulbs-page issues

kand:


kand:
[bulbs-page analytics issues](https://trello.com/c/2mEPaOLT/1359-bulbs-page-analytics-issues)

elements/bulbs-page/bulbs-page.js
```diff
@@ -14,10 +14,27 @@ import invariant from 'invariant';
...
 import './bulbs-page.scss';
 
-// if we scroll past multiple pages at once they will all trigger a
-// 'pagestart' event. We only want to fire one. We use this debouncer
-// to skip all but the last 'pagestart' handler
-let pageStartDebouncer = debouncePerFrame();
+let inViewRectDebouncer = debouncePerFrame();
+let inViewQueue = [];
+
+function drainInViewQueue () {
```
kand: @collin does this logic make sense? Are there any cases you can see this not working in?

collin: It seems sound to me.

We probably should get a confirmation from design that the urls are changing is a good place for UX.

kand: Apologies, missed this comment before merging. I talked with UX about this change and it should be all good.

collin: :+1:

PR #299, opened 2017-04-24
<bulbs-pdfobject>

jmelvnsn:
@collin @spra85 @theonion/front-end 

This is a handy little element created for the Trump Docs feature. It uses a small library called [pdfobject](https://pdfobject.com/) to force the browser to render a PDF inline. However, most mobile devices do not allow inline pdf embedding so I've added an attribute for a poster image that is a fallback in the case that a user's browser does not support inline rendering. 

```<bulbs-pdfobject src="https://s3.amazonaws.com/onionstatic/onion/custom/trump-docs/pdf-test.pdf" poster="http://images2.onionstatic.com/onion/5692/6/16x9/500.jpg"></bulbs-pdfobject>```

http://pdf-viewer.test.theonion.com/live/pdf-viewer-test-liveblog-55829/entry/275

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -97,6 +97,7 @@
     "node-sass": "^3.4.2",
     "object-map-to-array": "^1.0.0",
     "onionsay": "github:theonion/onionsay",
+    "pdfobject": "pipwerks/pdfobject#d37c492c026dc8abeac636d1d5d6eeebdb409e8b",
```
collin: Might be worth a comment here explaining this is intentionally pointing at a hash for use in production.

jmelvnsn: package.json don't like comments

collin: coo

elements/bulbs-pdfobject/bulbs-pdfobject-example.scss
```diff
@@ -0,0 +1,4 @@
+.pdfobject {
+  height: 500px!important;
+  width: 350px!important;
```
spra85: Do we want these both to be `!important` like this out of the gate?

spra85: Disregard.  Realized this is probably just for the example...

elements/bulbs-pdfobject/bulbs-pdfobject.js
```diff
@@ -0,0 +1,23 @@
...
+
+class BulbsPDFObject extends BulbsHTMLElement {
+  createdCallback () {
+    let poster = this.getAttribute('poster');
+    let options = {
+      fallbackLink: `<a href='[url]'><img src="${poster}"></a>`,
+    };
+    invariant(this.hasAttribute('poster'), '<bulbs-pdfobject> MUST have a \'poster\' attribute');
+    invariant(this.hasAttribute('src'), '<bulbs-pdfobject> MUST have a \'src\' attribute');
+    PDFObject.embed(this.getAttribute('src'), this, options);
```
brycedorn: actually, only comment I have is you could add an `inViewMonitor` to only load if in view. seems they're all being loaded on this `embed` call even though they're off-screen

collin: @brycedorn 👍 

talked to jamie about doing something like this:

```js
class PDFObject extends BulbsHTMLElement {
 createdCallback () {
    inViewMonitor.add(this);
    this.addEventListener('enterviewport', this.embedPDF.bind(this))
  }

  embedPDF () {
    PDFObject.embed(this.getAttribute('src'), this, options);
    inViewMonitor.remove(this);
  }
  
  detachedCallback () {
    inViewMonitor.remove(this);
  }
}
```

spra85: Makes sense @collin

PR #301, opened 2017-04-25
Fix inview monitor edge case

collin:
This catches a literal edge case (top edge of viewport) that jamie uncovered while testing the new `<bulbs-pdfobject>` element.

Rather than compare that the top of the element is `> 0px`, we allow for `>= 0px`.

Also updates the px values of some other tests to adapt to the new comparison.

Users should not notice this change.


lib/bulbs-elements/util/in-view-monitor.test.js
```diff
@@ -265,6 +265,16 @@ describe('InViewMonitor', () => {
...
   });
 
+  describe('isElementInViewport', () => {
+    it('zero height element at top of viewport is in view', () => {
+      el.style.height = '0px';
+      el.style.position = 'fixed';
+      el.style.top = '0px';
+
+      expect(InViewMonitor.isElementInViewport(el)).to.be.true;
+    });
```
collin: Here's the new test case.

collin:
@jmelvnsn I your new branch as the base for this, so if you approve it can merge straight into your branch.

PR #304, opened 2017-05-02
Patch for async load issue

jmelvnsn:
@collin @spra85


collin:
I think in the race condition this might have the effect of never calling `this.adsManager.asyncRefreshSlot(this);`.

If the `enterviewport` event fired before the adsManager was available this check wouldn't happen again unless the ad slot left and re-entered the viewport.

spra85:
@collin, I think that might be preferred for now as opposed to using the `invariant`

PR #308, opened 2017-05-19
Add some GA values to pdf-object fallback

jmelvnsn:
Will need to revisit to make `data-track-category` dynamic. But it is outside the scope of the request for Trump Docs GA events.  Since the PDFs have already been ingested in the CMS it would require updating all 80+ entries by passing the value in through the element. The end result of this work is it gives us some sort of metric with actual user interactions with the PDF. @theonion/front-end

### DIFF
--- a/elements/bulbs-pdfobject/bulbs-pdfobject.js
+++ b/elements/bulbs-pdfobject/bulbs-pdfobject.js
@@ -21,7 +21,7 @@ class BulbsPDFObject extends BulbsHTMLElement {
   embedPDF () {
     let poster = this.getAttribute('poster');
     let options = {
-      fallbackLink: `<a href='[url]' data-track-category="SF: Trump Docs 2017" data-track-action="Embed Fallback" data-track-label='[url]'><div class="fallback-btn">Tap to explore document</div><img src="${poster}"></a>`,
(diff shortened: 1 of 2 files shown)


jmelvnsn:
@brycedorn:

elements/bulbs-pdfobject/bulbs-pdfobject.js
```diff
@@ -21,7 +21,7 @@ class BulbsPDFObject extends BulbsHTMLElement {
   embedPDF () {
     let poster = this.getAttribute('poster');
     let options = {
-      fallbackLink: `<a href='[url]'><div class="fallback-btn">Tap to explore document</div><img src="${poster}"></a>`,
+      fallbackLink: `<a href='[url]' data-track-category="SF: Trump Docs 2017" data-track-action="Embed Fallback" data-track-label='[url]'><div class="fallback-btn">Tap to explore document</div><img src="${poster}"></a>`,
```
spra85: There's no way to pass in `SF: Trump Docs 2017` from the integrating app template?  Don't like hardcoding that

jmelvnsn: It's currently being used in liveblog which has it's own `data-track-category` value. So needs to be passed through the element to avoid bubbling up to the wrong category.  The issue is that this would need to be updated in 80+ places due to current Trump Docs being ingested. I'm fine with revisiting this  after launch.

PR #310, opened 2017-05-25
Fix issue where disable ads bool ignored

spra85:
## Why?

So the collapse/disable ads flag is not ignored

## How to test?

See instructions in the omni pr (https://github.com/theonion/omni/pull/1233)


elements/bulbs-reading-list/lib/reading-list-test-helper.js
```diff
@@ -1,50 +0,0 @@
...
-    item.appendChild(document.createElement('progress-bar'));
-    readingListArticles.appendChild(item);
-  });
-  readingList.appendChild(readingListMenu);
-  readingList.appendChild(readingListArticles);
-  flyoverMenu.appendChild(dfpAd);
-  flyoverMenu.appendChild(readingList);
-
-  return flyoverMenu;
-}
```
spra85: Not related but this is not even used anymore and it was causing the `webpack-dev-server` script command to bomb out

PR #311, opened 2017-06-06
Creative size

kjacovino:
**Purpose**
Add ability to pass in a VAST video size for the rail unit rather than have it hardcoded

**Ticket**
https://app.asana.com/0/342156387987716/354541582224232

@spra85 I'm not sure how to do like a test link, but I was testing it in the bulbs local webpack on the rail-player with campaign and it is working there..

Also not sure how the tests work yet, but figured I'd put this up here for now!

### DIFF
--- a/elements/bulbs-video/components/revealed.js
+++ b/elements/bulbs-video/components/revealed.js
@@ -55,6 +55,9 @@ export default class Revealed extends React.Component {
     let specialCoverage = this.props.targetSpecialCoverage || 'None';
     let autoplayInViewBool = typeof this.props.autoplayInView === 'string';
 
+    // Allowing creativeSize to be passed in in root.js
+    let creativeSize = this.props.creativeSize;
+
     let videoAdConfig = 'None';
     if (this.props.disableAds || this.props.video.disable_ads) {
       videoAdConfig = 'disable-ads';
@@ -78,6 +81,8 @@ export default class Revealed extends React.Component {
       BULBS_ELEMENTS_ONIONSTUDIOS_GA_ID,
       dimensions
     );
+
+
 
     // Making assignment copies here so we can mutate object structure.
     let videoMeta = Object.assign({}, this.props.video);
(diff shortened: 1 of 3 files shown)


elements/bulbs-video/components/revealed.js
```diff
@@ -55,6 +55,9 @@ export default class Revealed extends React.Component {
     let specialCoverage = this.props.targetSpecialCoverage || 'None';
     let autoplayInViewBool = typeof this.props.autoplayInView === 'string';
 
+    // Allowing creativeSize to be passed in in root.js
+    let creativeSize = this.props.creativeSize;
```
spra85: I would make this an optional attribute, with a default value of `640x480`, so simply change this to: 

```
let creativeSize = this.props.creativeSize || '640x480';
```

elements/bulbs-video/elements/rail-player/rail-player.js
```diff
@@ -60,6 +60,7 @@ Object.assign(RailPlayer, {
   },
   propTypes: {
     channel: PropTypes.string.isRequired,
+    creativeSize: PropTypes.string,
```
spra85: I don't think you actually need the `propType` defined here.

elements/bulbs-video/components/revealed.js
```diff
@@ -171,7 +176,7 @@ export default class Revealed extends React.Component {
     let type;
 
     // See docs (https://support.google.com/dfp_premium/answer/1068325?hl=en) for param info
-    baseUrl += '?sz=640x480';
+    baseUrl += `?sz=${this.props.creativeSize}`;
```
spra85: So, you should either find a way to use the `creativeSize` you defined above in the `let` assignment or do what you're doing here but otherwise the assignment above isn't used.

elements/bulbs-video/components/revealed.js
kjacovino: can do!

elements/bulbs-video/elements/rail-player/rail-player.js
kjacovino: true statement, still works without it

elements/bulbs-video/components/revealed.js
kjacovino: so if I removed the PropTypes line, does that mean this gets to stay?

PR #314, opened 2017-07-06
Third party build

imakeinterfaces:
This changeset allows third parties to use the ad-related bulbs-dfp component without react and other dependencies. 

- Splits up register.js into react and non-react dependent libraries.
- Removes analytics dependency from bulbs-dfp. 
- Requires InViewMonitor in bulbs-dfp as a standalone dependency instead of importing all utils.


spra85:
@imakeinterfaces, this is looking great as a start.  The split & naming convention of how you broke it out makes sense.  The only thing we have left to do, though, is convert webpack to use the new split files as part of its build step. 

If you run the following command, and look inside `/dist` locally, we'll want to come up with a varied strategy of providing a bundle with only register-element.js or register-react.js and their dependencies.  

```
./scripts/webpack-dev-build
```

You'll see after running that command as is, you'll get `vendor.bundle.js` as part of the distribution.  If a third-party was using the repo, you'd need that `vendor.bundle.js` file to be loaded before the `bulbs-dfp.js` file.

imakeinterfaces:
@spra85 I just added the changes to webpack.base as i had forgotten to commit that.  

Regarding the strategy for providing the bundle, the method i have used in mantle is to link directly to the required files like so: https://github.com/gawkermedia/kinja-mantle/blob/c6ac0758d64346160e8e6ab4c407ef544a47d674/packages/standalone-ad-manager/webpack.config.js#L7 Since I only need a very narrow subset of the components and utils i think this is a lot cleaner. Does that make sense?

spra85:
@imakeinterfaces, that makes a ton of sense.  Have you already confirmed the integration works fine with Mantle that way?  If so, can you send me the updated PR via Slack and then I'll review this build to make sure all looks well on the onion sites.

Also, can you work on getting the build to pass for this PR?  You'll want to run both the tests and the linter locally (both are failing): 

```
./scripts/test-js
./scripts/lint
```

collin:
Exciting PR 👍 💯

PR #321, opened 2017-09-28
this was not correctly using the autoplay attribute for the dimension

spra85:


elements/bulbs-video/components/revealed.js
```diff
@@ -68,7 +68,7 @@ export default class Revealed extends React.Component {
       'dimension5': hostChannel,
       'dimension6': specialCoverage,
       'dimension7': true, // 'has_player' from old embed
-      'dimension8': this.props.autoplay || autoplayInViewBool || 'None',
+      'dimension8': this.props.hasOwnProperty('autoplay') || autoplayInViewBool || 'None',
```
brycedorn: could you explain the reasoning behind this? I don't think props aren't ever [inherited](https://facebook.github.io/react/docs/composition-vs-inheritance.html#so-what-about-inheritance) so not sure  why `hasOwnProperty` would ever return false unless it's passed in

spra85: @brycedorn, it's mainly because the way it works now as a passed attribute, it's only ever autoplay if it's passed in at all.  

For example: 

```
<bulbs-video src="some url" autoplay> // This should pass in `dimension8` as true
```

```
<bulbs-video src="some url"> // This should pass in `dimension8` as None
```

PR #324, opened 2018-03-09
Quiz and Clickventure embeds resize (Kinja) parent frame

mparent61:
Used for Clickhole migration to Kinja.

Leverages Kinja Mantle's built-in support for passing resize messages from within an embed iframe.

### DIFF
--- a/elements/bulbs-clickventure/bulbs-clickventure.js
+++ b/elements/bulbs-clickventure/bulbs-clickventure.js
@@ -20,7 +20,7 @@ class BulbsClickventure extends BulbsHTMLElement {
     $element.data('clickventurePlugin', new Clickventure($element));
     $element.on('clickventure-page-change', fireAnalytics);
 
-    // Embeded CVs can resize their parent frame
+    // Embedded CVs can resize their parent frame
     resizeParentFrame(document.body.clientHeight);
   }
 }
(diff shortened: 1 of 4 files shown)


mparent61:
@spra85 -- still need to test via temp release, but is this what you had in mind?

elements/bulbs-clickventure/bulbs-clickventure.js
```diff
@@ -18,6 +19,9 @@ class BulbsClickventure extends BulbsHTMLElement {
     let $element = $(this);
     $element.data('clickventurePlugin', new Clickventure($element));
     $element.on('clickventure-page-change', fireAnalytics);
+
+    // Embeded CVs can resize their parent frame
+    resizeParentFrame(document.getElementsByTagName('body')[0].clientHeight);
```
collin: Looks nice, can use the shortcut `document.body` here to get the body element.

mparent61: Thank you kind sir!

lib/bulbs-elements/util/resize-parent-frame.js
```diff
@@ -0,0 +1,15 @@
+// Resize a (Kinja) parent frame, using Kinja's parent notification support.
+
+export default function resizeParentFrame (height) {
+  if (window.parent !== null) {
```
spra85: Should be able to just use: 

```
if (window.parent) {
```

PR #332, opened 2018-03-28
Set data-src and class on splitpic images

MichaelButkovic:
This sets the data-src and class on splitpic images, which will then be rendered by Cloudinary.


elements/bulbs-splitpic/bulbs-splitpic.test.js
```diff
@@ -9,5 +9,6 @@ describe('<bulbs-splitpic>', function () {
 
   it('renders an <bulbs-splitpic>', function () {
     assert.equal(element.tagName.toLowerCase(), 'bulbs-splitpic');
+    debugger;
```
spra85: Remove `debugger`

